PR #4, opened 2021-11-02
Fix command `hanami db rollback`

pauc:
The command `hanami db rollback` wasn't working as expected: instead of rolling back the last migration it was crashing with a `NameError: uninitialized constant Hanami::CLI::Commands::DB::Utils::Database::Sequel`

Once this exception was fixed I also changed the logic to select the target migration, otherwise the last migration is selected and the database remains unchanged.

### TODO

- [ ] Still failing when there is only one migration applied

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,8 +8,8 @@ unless ENV["CI"]
   gem "yard",   require: false
 end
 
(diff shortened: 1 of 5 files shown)


jodosha:
@pauc Thanks for this PR, we'll add again `hanami db` commands soon.

PR #5, opened 2021-12-04
Fix console context (make slice readers properly available)

timriley:
This PR makes the slice readers properly available in the console context methods. With the previous approach, they were unavailable (e.g. in the console for our application template, typing `main` would raise a `NoMethodError`).

This could have been fixed most simply by changing `extend(Plugins::SliceReaders.new(app))` inside `Context.extended` so it was actually `other.extend(Plugins::SliceReaders.new(app))`. However, I took this change to refactor this module since we didn't actually have to define all this behavior at the moment of being extended (`other` was never used anywhere!) — since we have all the information we need at the time of initialization, that's the best time to define all the behavior. We can leave the module hooks alone until we actually need access to that `other` object for some reason.

Along with this, I've changed the slice readers so that they no longer automatic delegate to top-level components based on method_missing. This is a neat trick but I think it may lead to more confusion than convenience, especially given that the majority of slice behavior will not be available via these shortcuts, since most slice behavior will not typically reside in top-level components.

Lastly, I've also added a couple of extra top-level console convenience methods: `app` and `application`, both returning the current Hanami application. In actuality, `self` inside the console is _kind of_ a way to get to the application, but this is non-obvious, and its still somewhat indirect, since it won't return the application itself, but rather the REPL's context object that just so happens to delegate its missing methods to the application.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ unless ENV["CI"]
 end
 
 gem "hanami", require: false, git: "https://github.com/hanami/hanami.git", branch: "main"
+gem "hanami-router", require: false, git: "https://github.com/hanami/router.git", branch: "main"
 gem "hanami-view", require: false, git: "https://github.com/hanami/view.git", branch: "main"
 
 gem "rack"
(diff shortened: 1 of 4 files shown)


timriley:
@solnic Flagging you in particular for review here since you worked on this structure :)

jodosha (approved):

solnic (approved):

PR #12, opened 2022-02-15
Add `hanami server` command

waiting-for-dev:
The server is just a thin wrapper on top of Rack::Server. The options
that it accepts fall into two different categories:

- When not explicitly set, port and host are not passed to the rack
server instance. This way, they can be configured through the configured
rack handler (e.g., the puma configuration file).

- All others are always given by the Hanami command.

The test for the command forks a new process so that the one running the
tests is not blocked and can assert that endpoint is working. Standard
outputs are silenced to avoid the command output being rendered.

### DIFF
--- a/lib/hanami/cli/commands/monolith/server.rb
+++ b/lib/hanami/cli/commands/monolith/server.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+require "rack"
+require_relative "../application"
+require_relative "../../server"
+
+module Hanami
+  module CLI
+    module Commands
+      module Monolith
+        # Launch Hanami web server.
+        #
+        # It's intended to be used only on development. For production, you
+        # should use the rack handler command directly (i.e. `bundle exec puma
+        # -C config/puma.rb`).
+        #
+        # The server is just a thin wrapper on top of Rack::Server. The options that it
+        # accepts fall into two different categories:
+        #
+        # - When not explicitly set, port and host are not passed to the rack
+        # server instance. This way, they can be configured through the
+        # configured rack handler (e.g., the puma configuration file).
+        #
+        # - All others are always given by the Hanami command.
+        #
+        # Run `bundle exec hanami server -h` to see all the supported options.
+        class Server < Application
+          desc "Start Hanami server"
+
+          option :host, default: nil, required: false, desc: "The host address to bind to (falls back to the rack handler)"
+          option :port, default: nil, required: false, desc: "The port to run the server on (falls back to the rack handler)"
+          option :config, default: 'config.ru', required: false, desc: "Rack configuration file"
+          option :debug, default: false, required: false, desc: "Turn on/off debug output", type: :boolean
+          option :warn, default: false, required: false, desc: "Turn on/off warnings", type: :boolean
+
+          private attr_reader :server
+
+          def initialize(server: Hanami::CLI::Server.new)
+            @server = server
+          end
+
+          # @api private
+          def call(...)
+            server.call(...)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/hanami/cli/server.rb
+++ b/lib/hanami/cli/server.rb
@@ -1,0 +1,43 @@
+module Hanami
+  module CLI
+    # @api private
+    class Server
+      attr_reader :rack_server
+
+      RACK_FALLBACK_OPTIONS = {
+        host: :Host,
+        port: :Port
+      }.freeze
+
+      OVERRIDING_OPTIONS = {
+        config: :config,
+        debug: :debug,
+        warn: :warn
+      }.freeze
+
+      def initialize(rack_server: Rack::Server)
+        @rack_server = rack_server
+      end
+
+      def call(**options)
+        rack_server.start(Hash[
+          extract_rack_fallback_options(options) + extract_overriding_options(options)
+        ])
+      end
+
+      private
+
+      def extract_rack_fallback_options(options)
+        RACK_FALLBACK_OPTIONS.filter_map do |(name, rack_name)|
+          options[name] && [rack_name, options[name]]
+        end
+      end
+
+      def extract_overriding_options(options)
+        OVERRIDING_OPTIONS.map do |(name, rack_name)|
+          [rack_name, options[name]]
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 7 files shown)


lib/hanami/cli/commands/monolith/server.rb
```diff
@@ -0,0 +1,42 @@
...
+          option :port, default: nil, required: false, desc: "The port to run the server on (falls back to the rack handler)"
+          option :config, default: 'config.ru', required: false, desc: "Rack configuration file"
+          option :daemonize, default: false, required: false, desc: "Daemonize the server"
+          option :debug, default: false, required: false, desc: "Turn on/off debug output"
+          option :pid, default: nil, required: false, desc: "Path to write a pid file after daemonize"
+          option :warn, default: false, required: false, desc: "Turn on/off warnings"
+
+          # @api private
+          def call(**options)
+            Hanami::CLI::Server.new.call(**options)
```
jodosha, line 36: @waiting-for-dev Just a stylistic feedback regarding showcasing `dry-cli`.

```ruby
private attr_reader :server

def initialize(server: Hanami::CLI::Server.new)
  @server = server
end

def call(...)
  server.call(...)
end
```

lib/hanami/cli/server.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      OVERRIDING_OPTIONS = {
+        config: :config,
+        daemonize: :daemonize,
+        debug: :debug,
+        pid: :pid,
+        warn: :warn
+      }.freeze
+
+      def initialize(rack_server = Rack::Server)
```
jodosha, line 20:
```suggestion
      def initialize(rack_server: Rack::Server)
```

lib/hanami/cli/server.rb
```diff
@@ -0,0 +1,45 @@
+module Hanami
+  module CLI
+    # @api private
+    class Server
+      attr_reader :rack_server
```
jodosha, line 5: @waiting-for-dev Please move this below constants. Also, please make it private `private attr_reader :rack_server`

jodosha (changes requested):
@waiting-for-dev Thanks for this PR, apart from the minor inline comments, I spotted a few problems by trying this PR with https://github.com/hanami/hanami-2-application-template.

## Host

```
bundle exec server --host=127.0.0.1
```

It binds to `127.0.0.1` host (correct), but it doesn't respect the port. Expected value is `2300` (default Hanami port), but it uses `9292`.

## Daemonize (true)

```
bundle exec hanami server --daemonize=true
```

It will start the background process (good), but on the wrong host and port. Expected `0.0.0.0:2300`, current is `localhost:9292`.

## Daemonize (false)

```
bundle exec hanami server --daemonize=false
```

It will start the background process (not expected).

## Debug (false)

```
bundle exec hanami server --debug=false
```

It will debug the app bootstrap (not expected).

## Warn (false)

```
bundle exec hanami server --warn=false
```

It will print the Ruby warnings (not expected).

## Forwarding CLI arguments

Given we can't map all the CLI arguments for Rack servers, it's important to forward them. `dry-cli` supports this feature.

```
mv config/puma.rb config/tuna.rb
bundle exec hanami server -- -C config/tuna.rb
```

Expected: to pass extra CLI arguments to the underlying server.

lib/hanami/cli/commands/monolith/server.rb
waiting-for-dev: Thanks for the suggestion. Done!

lib/hanami/cli/server.rb
waiting-for-dev: Done!

lib/hanami/cli/server.rb
waiting-for-dev: We're leveraging the public visibility of `:rack_server` to mock it in the test suite. Besides, I think it makes sense to mark the whole class as private for the user. The public interface is the command but not the server. WDYT?

waiting-for-dev:
> ## Daemonize (true)
> 
> ```
> bundle exec hanami server --daemonize=true
> ```
> 
> It will start the background process (good), but on the wrong host and port. Expected `0.0.0.0:2300`, current is `localhost:9292`

It turns out to be an issue on rack. When daemonizing, the working directory is changed to `/`. Because of that, puma can't find its config file later on. Please, take a look at the [issue I opened in Rack](https://github.com/rack/rack/issues/1813).

Like with the host + port issue [commented above](https://github.com/hanami/cli/pull/12#issuecomment-1041875697), that's a problem that has been unnoticed on Rails because both it and the puma config file use the same defaults for host and port. In our case, I think it's wiser not to set any defaults and let the rack handler set them. This way we avoid consulted logic (see https://schneems.com/2017/03/13/puma-ports-and-polish/).

> Forwarding CLI arguments
> 
> Given we can't map all the CLI arguments for Rack servers, it's important to forward them. dry-cli supports this feature.
> 
> mv config/puma.rb config/tuna.rb
> bundle exec hanami server -- -C config/tuna.rb
> 
> Expected: to pass extra CLI arguments to the underlying server.

Unfortunately, it won't work as expected. We're wrapping the rack server, but not the underlying handler. So, we can't forward to the rack handler CLI command. I don't think that's too bad, as `hanami server` should only be used in development (see [comment on puma repo](https://github.com/puma/puma/issues/512#issuecomment-77800102) for details). On top of that, different rack handlers could understand different formatting for arguments, so we should create adapters for each one, which I don't think it's worth the effort. What I'd do is forward the variadic arguments to the rack handler. WDYT?

Please, @jodosha, see also my comments to your suggestions above.

BTW, you can see a failing test, which won't work on GH actions probably because TCP connections are blocked. I'm not very familiar with GH actions, though. Any suggestion? I think it's good to have such an integration test, so we could also have it run only locally (controlling with an env var).

We're setting `pid` with `nil` as default. I've thought that maybe I could follow up with `hanami server stop` & `hanami server restart` command and work with the pid file then. What do you think about having those types of commands?

waiting-for-dev:
> It turns out to be an issue on rack. When daemonizing, the working directory is changed to /. Because of that, puma can't find its config file later on. Please, take a look at the https://github.com/rack/rack/issues/1813.

After discussion on the referenced issue and internal talks, we've decided not to support daemonization at this point. The value it provides in development is not clear due to its drawbacks (lack of stdout/stderr, cumbersome to stop/restart a server...). We might reconsider in the future.

I've added a commit removing both `daemonization` & `pid` option (which now makes no sense).

PR #19, opened 2022-05-24
Prefer NoMethodError over NotImplementedError

tak1n:
Relates to https://github.com/hanami/controller/pull/377


jodosha:
@tak1n Thanks for this PR, but in the end I had to use a subclass of `Hanami::CLI::Error` to avoid exceptions to bubble up and print the stack trace in the shell. See https://github.com/hanami/cli/pull/63

PR #22, opened 2022-06-27
Add `hanami server` command

waiting-for-dev:
The server is just a thin wrapper on top of Rack::Server. The options
that it accepts fall into two different categories:

- When not explicitly set, port and host are not passed to the rack
server instance. This way, they can be configured through the configured
rack handler (e.g., the puma configuration file).

- All others are always given by the Hanami command.

The test for the command forks a new process so that the one running the
tests is not blocked and can assert that endpoint is working. Standard
outputs are silenced to avoid the command output being rendered.

Replaces #12

See https://github.com/puma/puma/pull/2832 for issue on Puma

### DIFF
--- a/hanami-cli.gemspec
+++ b/hanami-cli.gemspec
@@ -38,4 +38,5 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "rspec", "~> 3.9"
   spec.add_development_dependency "rubocop", "~> 1.0"
+  spec.add_development_dependency "puma"
 end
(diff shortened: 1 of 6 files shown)


jodosha (approved):
@waiting-for-dev LGTM, thanks!

---

Note for the future: `Hanami::CLI::Server` depends on `Rack::Server`, but `hanami-cli` doesn't depend on `rack`.

This works because we assume the presence of `hanami`.

This makes an ciclic dependency:

* Explicit dependency `hanami` -> `hanami-cli`: because Hanami has `hanami-cli` in its gemspec
* Implicit dependency `hanami-cli` -> `hanami`: because certain CLI commands won't work without Hanami.

PR #23, opened 2022-06-29
Register `hanami server` command

jodosha:
# Enhancements

* Register `hanami server` command
* Make Rubocop happy

---

Follow up of https://github.com/hanami/cli/pull/22

PR #24, opened 2022-06-29
Set default `hanami server` port to `2300`

jodosha:
Follow up of #22 and #23


lib/hanami/cli/commands/application/server.rb
```diff
@@ -41,8 +44,8 @@ def initialize(server: Hanami::CLI::Server.new)
           end
 
           # @api private
-          def call(...)
-            server.call(...)
+          def call(port: DEFAULT_PORT, **kwargs)
```
waiting-for-dev, line 47: @jodosha, do we need it? We're already setting the default in the `option` definition.

PR #26, opened 2022-07-12
Add `hanami routes` command

waiting-for-dev:
Without the `slice` option, it will inspect the application routes:

```ruby
$ hanami routes
```

All the formatters available from `hanami-router` are available:

```ruby
$ hanami routes --format=csv
```

A custom formatter registered in the application container can also be
used. It needs to be identified by its key:

```ruby
$ hanami routes --format=custom_routes_formatter
```

### DIFF
--- a/lib/hanami/cli/commands/app/routes.rb
+++ b/lib/hanami/cli/commands/app/routes.rb
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+require "hanami"
+require "hanami/router/inspector"
+
+module Hanami
+  module CLI
+    module Commands
+      module App
+        # Inspect the application routes
+        #
+        # All the formatters available from `hanami-router` are available:
+        #
+        # ```
+        # $ hanami routes --format=csv
+        # ```
+        #
+        # Experimental: You can also use a custom formatter registered in the
+        # application container. You can identify it by its key:
+        #
+        # ```
+        # $ hanami routes --format=custom_routes_formatter
+        # ```
+        class Routes < Hanami::CLI::Command
+          DEFAULT_FORMAT = "human_friendly"
+          private_constant :DEFAULT_FORMAT
+
+          VALID_FORMATS = [
+            DEFAULT_FORMAT,
+            "csv"
+          ].freeze
+          private_constant :VALID_FORMATS
+
+          desc "Inspect application"
+
+          option :format,
+                 default: DEFAULT_FORMAT,
+                 required: false,
+                 desc: "Output format"
+
+          # @api private
+          def call(format: DEFAULT_FORMAT, slice: nil)
+            require "hanami/prepare"
+            inspector = Hanami::Router::Inspector.new(formatter: resolve_formatter(format))
+            app.router(inspector: inspector)
+            out.puts inspector.call
+          end
+
+          private
+
+          def resolve_formatter(format)
+            if VALID_FORMATS.include?(format)
+              resolve_formatter_from_hanami_router(format)
+            else
+              resolve_formatter_from_app(format)
+            end
+          end
+
+          def resolve_formatter_from_hanami_router(format)
+            case format
+            when "human_friendly"
+              require "hanami/router/formatter/human_friendly"
+              Hanami::Router::Formatter::HumanFriendly.new
+            when "csv"
+              require "hanami/router/formatter/csv"
+              Hanami::Router::Formatter::CSV.new
+            end
+          end
+
+          # Experimental
+          def resolve_formatter_from_app(format)
+            app[format]
+          end
+
+          def app
+            Hanami.app
+          end
+        end
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -8,8 +8,8 @@ unless ENV["CI"]
   gem "yard",   require: false
 end
 
-gem "hanami", require: false, git: "https://github.com/hanami/hanami.git", branch: "main"
-gem "hanami-router", require: false, git: "https://github.com/hanami/router.git", branch: "main"
+gem "hanami", require: false, git: "https://github.com/hanami/hanami.git", branch: "waiting-for-dev/inspector"
+gem "hanami-router", github: "hanami/router", branch: :main
 
 gem "rack"
 
--- a/lib/hanami/cli/commands/app.rb
+++ b/lib/hanami/cli/commands/app.rb
@@ -8,6 +8,7 @@ module Hanami
         require_relative "app/install"
         require_relative "app/console"
         require_relative "app/server"
+        require_relative "app/routes"
         # require_relative "app/generate"
         # require_relative "app/db/create"
         # require_relative "app/db/create_migration"
@@ -27,6 +28,7 @@ module Hanami
             register "install", Commands::App::Install
             register "console", Commands::App::Console, aliases: ["c"]
             register "server",  Commands::App::Server,  aliases: ["s"]
+            register "routes",  Commands::App::Routes
 
             # FIXME: temporary disabled for Hanami v2.0.0.alpha2
             # register "install", Install
--- a/spec/fixtures/test/config/routes.rb
+++ b/spec/fixtures/test/config/routes.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+require "hanami/routes"
+
+module Test
+  class Routes < Hanami::Routes
+    define do
+      get "/", to: "home.index"
+      get "/about", to: "home.about"
+    end
+  end
+end
--- a/spec/unit/hanami/cli/commands/app/routes_spec.rb
+++ b/spec/unit/hanami/cli/commands/app/routes_spec.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require "hanami/cli/commands/app/routes"
+
+RSpec.describe Hanami::CLI::Commands::App::Routes, :app, :command do # see fixture app for the defined routes
+  # TODO: better Hanami tear down
+  after do
+    app = Hanami.app
+    app.remove_instance_variable(:@_router) if app.instance_variable_defined?(:@_router)
+  end
+
+  it "defaults to the human friendly formatter" do
+    command.call
+
+    expect(output).to match %r{
+     ^GET\s+/\s+home\.index.*
+      GET\s+/about\s+home\.about\s+$
+    }xm
+  end
+
+  it "can use the csv formatter" do
+    command.call(format: "csv")
+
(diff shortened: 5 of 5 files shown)


lib/hanami/cli/commands/app/routes.rb
```diff
@@ -0,0 +1,92 @@
...
+
+          def resolve_formatter_from_app(format)
+            app[format]
+          end
+
+          def resolve_slice(slice)
+            slice ? app.slices[slice.to_sym] : app
+          end
+
+          def app
```
jodosha, line 85: @waiting-for-dev Could you please move this to `#initialize`?

```ruby
def initialize(app: Hanami.app, **)
  super
  @app = app
end

private

attr_reader :app
```

jodosha (commented):
@waiting-for-dev The command isn't registered 😂

lib/hanami/cli/commands/app/routes.rb
```diff
@@ -0,0 +1,92 @@
...
+
+            file
+              .then { |f| inflector.classify(f) }
+              .then { |class_name| inflector.constantize(class_name) }
+              .then(&:new)
+          rescue LoadError
+            false
+          end
+
+          def resolve_formatter_from_app(format)
```
jodosha, line 77: @waiting-for-dev Given we're providing Human friendly and CSV, I wouldn't bother to let final users to specify a formatter.

Technical reasons are:

* One more feature to support
* We need to document where to put this file and how it should be structured. I assume that we would provide a superclass for it.
* In case of changes in on the superclass, we'll need to support a migration.

The cost/benefit analysis for me is negative: too much burden for a very little gain in terms of features.

Could you please remove it?

lib/hanami/cli/commands/app/routes.rb
```diff
@@ -0,0 +1,92 @@
...
+
+          private
+
+          def resolve_formatter(format)
+            resolve_formatter_from_hanami_router(format) ||
+              resolve_formatter_from_app(format)
+          end
+
+          def resolve_formatter_from_hanami_router(format)
+            file = "hanami/router/formatter/#{format}"
```
jodosha, line 66: @waiting-for-dev Requiring a file without filtering the input against a whitelist can be a security threat.

jodosha (changes requested):
@waiting-for-dev The command crashes when testing it manually:

```
$ bundle exec hanami routes
bundler: failed to load command: hanami (/Users/jodosha/.gem/ruby/3.1.0/bin/hanami)
/Users/jodosha/.gem/ruby/3.1.0/bundler/gems/dry-container-5afce79243b9/lib/dry/container/resolver.rb:31:in `block in call': key not found: "human_friendly" (Dry::Container::KeyError)
	from /Users/jodosha/.gem/ruby/3.1.0/bundler/gems/dry-container-5afce79243b9/lib/dry/container/resolver.rb:27:in `fetch'
	from /Users/jodosha/.gem/ruby/3.1.0/bundler/gems/dry-container-5afce79243b9/lib/dry/container/resolver.rb:27:in `call'
	from /Users/jodosha/.gem/ruby/3.1.0/bundler/gems/dry-container-5afce79243b9/lib/dry/container/mixin.rb:171:in `resolve'
	from /Users/jodosha/.gem/ruby/3.1.0/bundler/gems/dry-system-de908f554916/lib/dry/system/container.rb:548:in `resolve'
	from /Users/jodosha/.gem/ruby/3.1.0/bundler/gems/dry-container-5afce79243b9/lib/dry/container/mixin.rb:184:in `[]'
	from /Users/jodosha/.gem/ruby/3.1.0/bundler/gems/hanami-8dd6e641c104/lib/hanami/slice.rb:152:in `[]'
	from /Users/jodosha/Code/hanami/cli/lib/hanami/cli/commands/app/routes.rb:78:in `resolve_formatter_from_app'
	from /Users/jodosha/Code/hanami/cli/lib/hanami/cli/commands/app/routes.rb:62:in `resolve_formatter'
	from /Users/jodosha/Code/hanami/cli/lib/hanami/cli/commands/app/routes.rb:53:in `call'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/dry-cli-0.7.0/lib/dry/cli.rb:116:in `perform_registry'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/dry-cli-0.7.0/lib/dry/cli.rb:65:in `call'
	from /Users/jodosha/Code/hanami/cli/exe/hanami:10:in `<top (required)>'
	from /Users/jodosha/.gem/ruby/3.1.0/bin/hanami:25:in `load'
	from /Users/jodosha/.gem/ruby/3.1.0/bin/hanami:25:in `<top (required)>'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/cli/exec.rb:58:in `load'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/cli/exec.rb:58:in `kernel_load'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/cli/exec.rb:23:in `run'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/cli.rb:479:in `exec'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/cli.rb:31:in `dispatch'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/cli.rb:25:in `start'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/exe/bundle:48:in `block in <top (required)>'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/lib/bundler/friendly_errors.rb:117:in `with_friendly_errors'
	from /Users/jodosha/.gem/ruby/3.1.0/gems/bundler-2.3.16/exe/bundle:36:in `<top (required)>'
	from /Users/jodosha/.gem/ruby/3.1.0/bin/bundle:25:in `load'
	from /Users/jodosha/.gem/ruby/3.1.0/bin/bundle:25:in `<main>'
```

lib/hanami/cli/commands/app/routes.rb
waiting-for-dev: Hey, @jodosha,

IMHO, that would be a beneficial feature. It'd give the option to integrate easily formatters which output could be imported into standard developer tools like Postman or OpenAPI.

> We need to document where to put this file and how it should be structured. I assume that we would provide a superclass for it.

No need for a superclass. It only needs to be [called by the inspector with the routes](https://github.com/hanami/router/blob/9c2b746d41725b3c2e511cf936db0b4ed32e76ee/lib/hanami/router/inspector.rb#L36). Yeah, of course, we need documentation, but like with everything else :slightly_smiling_face: 

> In case of changes in on the superclass, we'll need to support a migration.
No superclass. Of course, the signature of the `#call` method could change at some point. But, as long as we obey with semver, it's like any other change in Hanami. I don't see why we need to support migration.

Do you still think it's better to remove it?

lib/hanami/cli/commands/app/routes.rb
waiting-for-dev: @jodosha, at that point, `Hanami.app` isn't prepared yet, so [it'll raise an error](https://github.com/hanami/hanami/blob/70eed47749e5df602f5749d77dbc529d431f5602/lib/hanami.rb#L15).

waiting-for-dev:
> @waiting-for-dev The command crashes when testing it manually:

It should be fixed now :pray:

jodosha (changes requested):
@waiting-for-dev Thanks for making it to work again.

There is still a problem with block routes.

Example:

```ruby
      get "/block" do
        "Block"
      end
```

```shell
⚡ be hanami routes
GET     /block                        NilClass
```

It prints `NilClass`, I would go with `(block)` to mimic `(proc)`.

lib/hanami/cli/commands/app/routes.rb
waiting-for-dev: @jodosha, TBH, after all the recent changes in Hanami, I'm not sure where should it go, but it should be the standard Hanami directory for "lib" code (`lib/`, `config/providers/` ? cc @timriley)

waiting-for-dev:
> There is still a problem with block routes.

Fixed in https://github.com/hanami/router/pull/224

jodosha (approved):

lib/hanami/cli/commands/app/routes.rb
timriley: I haven't been following this PR and I can't catch up tonight, so all I can do at this point is answer your question based on the surrounding code. So: it looks like you want the formatter to be a component registered in the Hanami app container. In this case it should go into `app/` (or if it is a complex component that requires its own configuration before it can be registered, etc., then yes, you could make a provider for it).

waiting-for-dev: Thanks, @timriley! Thoughts, @jodosha? :slightly_smiling_face:

lib/hanami/cli/commands/app/routes.rb
```diff
@@ -0,0 +1,95 @@
...
+          def call(format: DEFAULT_FORMAT, slice: nil)
+            require "hanami/prepare"
+            inspector = Hanami::Router::Inspector.new(formatter: resolve_formatter(format))
+            app.router(inspector: inspector)
+            out.puts inspector.call
+          end
+
+          private
+
+          attr_reader :app
```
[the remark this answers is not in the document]
waiting-for-dev: Thanks, it was a left-over. Removed.

lib/hanami/cli/commands/app/routes.rb
waiting-for-dev: After talking about it off-line, we've decided to support it in an experimental phase.

PR #29, opened 2022-07-21
Tweaks to generated app

timriley:
Just took a moment to run through the app we're generating, and here are a few tweaks. Should be easy to look through them commit by commit.

This PR also relates to https://github.com/hanami/hanami/pull/1186, because after this is merged, we'll need to ensure dry-types is in our generated app `Gemfile`.


cllns:
What do you think about adding a `slices/.keep` file while we're at it? It's a core part of the framework and this reflects its status as a first-class concept. We have a `lib/tasks/.keep` which is less 'core' IMO.

Even if people remove the folder because it's empty, at least they'll see it right after generating a new app, to give them some familiarity with the term, even if they don't know what it means.

jodosha:
@cllns `slices/` is generated by `hanami g slice foo`, I wouldn't bother to add `slices/.keep`.

For `lib/tasks` that's a hint for developers about where to put Rake tasks.

lib/hanami/cli/generators/gem/app/gemfile.erb
```diff
@@ -2,14 +2,15 @@
...
 gem "hanami-router", "<%= hanami_version %>"
 gem "hanami-controller", "<%= hanami_version %>"
-gem "hanami", "<%= hanami_version %>"
 
+gem "dry-types"
 gem "puma"
+gem "rake"
 
 group :cli, :development, :test do
+  gem "dotenv"
```
jodosha, line 14: @timriley `:cli` group is only for gems that enhance `hanami-cli`.

This should be:

```ruby
group :development, :test do
  gem "dotenv"
end
```

lib/hanami/cli/generators/gem/app/types.erb
```diff
@@ -1,10 +1,11 @@
-# auto_register: false
```
jodosha, line 1: @timriley Do we expect this to be auto-registered in app container?

jodosha (changes requested):
@timriley Please check my comment about `Gemfile` groups 😉

lib/hanami/cli/generators/gem/app/gemfile.erb
timriley: Oh thanks, I've addressed this now! (Via a force push)

lib/hanami/cli/generators/gem/app/types.erb
timriley: Nope, that won't be a concern, since our generator is writing it into `lib/`, not `app/`:

```ruby
fs.write("lib/#{app}/types.rb", t("types.erb", context))
```

`lib/` isn't touched by our container auto-registration at all, so we don't need this comment 😄

timriley:
@jodosha This is ready for you to check again, have addressed your feedback!

jodosha (approved):

PR #31, opened 2022-07-23
Restore hanami-validations

swilgosz:
Resolves: https://trello.com/c/Lx86f21b/192-hanami-validations-not-included-in-a-brand-new-application

When you create a new hanami app, using `hanami new`, the `hanami-validations` is not included. This results in a problem, where we cannot use

```ruby
params do
end
```

within actions - I believe this should be built-in feature.

### DIFF
--- a/lib/hanami/cli/generators/gem/app/gemfile.erb
+++ b/lib/hanami/cli/generators/gem/app/gemfile.erb
@@ -6,6 +6,7 @@ gem "rake"
 
 gem "hanami-router", "<%= hanami_version %>"
 gem "hanami-controller", "<%= hanami_version %>"
+gem "hanami-validations", "<%=hanami_version %>"
 gem "hanami", "<%= hanami_version %>"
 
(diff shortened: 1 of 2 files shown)


swilgosz:
QA result:

image

lib/hanami/cli/generators/gem/app/gemfile.erb
```diff
@@ -6,6 +6,7 @@ gem "rake"
 
 gem "hanami-router", "<%= hanami_version %>"
 gem "hanami-controller", "<%= hanami_version %>"
+gem "hanami-validations", "<%=hanami_version %>"
```
jodosha, line 9:
```suggestion
gem "hanami-validations", "<%= hanami_version %>"
```

jodosha (approved):
@swilgosz Thanks for taking care of this.

Manually tested in a Hanami app that `hanami-validations` work.

PR #32, opened 2022-07-27
Introduce `hanami generate slice`

jodosha:
# Feature

Add `hanami generate slice` CLI command to Hanami applications.

It performs the following actions:

  * Edit `config/routes.rb` to add a `slice` block
  * Create `slices/[slice]` directory
  * Create `slices/[slice]/action.rb` file (with `MySlice::Action < MyApp::Action` class definition)
  * Create `slices/[slice]/actions/.keep` file

Example:

```shell
⚡ bundle exec hanami generate slice admin
⚡ tree slices
slices
└── admin
    ├── action.rb
    └── actions

2 directories, 1 file
```

### DIFF
--- a/lib/hanami/cli/generators/app/slice.rb
+++ b/lib/hanami/cli/generators/app/slice.rb
@@ -14,23 +14,24 @@ module Hanami
             @inflector = inflector
           end
 
-          def call(app, slice, slice_url_prefix, context: SliceContext.new(inflector, app, slice, slice_url_prefix)) # rubocop:disable Metrics/AbcSize
+          def call(app, slice, slice_url_prefix, context: SliceContext.new(inflector, app, slice, slice_url_prefix))
             fs.inject_line_before_last(fs.join("config", "routes.rb"), /end/, t("routes.erb", context).chomp)
 
             fs.mkdir(directory = "slices/#{slice}")
 
             fs.chdir(directory) do
+              fs.write("slice.rb", t("slice.erb", context))
               fs.write("action.rb", t("action.erb", context))
-              fs.write("view.rb", t("view.erb", context))
-              fs.write("entities.rb", t("entities.erb", context))
-              fs.write("repository.rb", t("repository.erb", context))
+              # fs.write("view.rb", t("view.erb", context))
+              # fs.write("entities.rb", t("entities.erb", context))
+              # fs.write("repository.rb", t("repository.erb", context))
 
               fs.write("actions/.keep", t("keep.erb", context))
-              fs.write("views/.keep", t("keep.erb", context))
-              fs.write("templates/.keep", t("keep.erb", context))
-              fs.write("templates/layouts/.keep", t("keep.erb", context))
-              fs.write("entities/.keep", t("keep.erb", context))
-              fs.write("repositories/.keep", t("keep.erb", context))
+              # fs.write("views/.keep", t("keep.erb", context))
+              # fs.write("templates/.keep", t("keep.erb", context))
+              # fs.write("templates/layouts/.keep", t("keep.erb", context))
+              # fs.write("entities/.keep", t("keep.erb", context))
+              # fs.write("repositories/.keep", t("keep.erb", context))
             end
           end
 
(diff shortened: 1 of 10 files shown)


lib/hanami/cli/generators/app/slice.rb
```diff
@@ -14,23 +14,24 @@ def initialize(fs:, inflector:)
             @inflector = inflector
           end
 
-          def call(app, slice, slice_url_prefix, context: SliceContext.new(inflector, app, slice, slice_url_prefix)) # rubocop:disable Metrics/AbcSize
-            fs.inject_line_before_last(fs.join("config", "routes.rb"), /end/, t("routes.erb", context).chomp)
+          def call(app, slice, slice_url_prefix, context: SliceContext.new(inflector, app, slice, slice_url_prefix))
+            fs.inject_line_at_block_bottom(fs.join("config", "routes.rb"), /define/, t("routes.erb", context).chomp)
```
timriley, line 18: Since the `slice_url_prefix` is optional, should we make this conditional on `slice_url_prefix` being truthy?

UPDATE: I've looked through the code a bit more and I see in `Commands::App::Generate::Slice` that we actually default the slice prefix to `"/"` if it isn't otherwise specified. I wonder if it is actually helpful to let the nil go all the way through so we can skip adding routes sections if slices are intended to be "headless"? What do you think?

lib/hanami/cli/generators/app/slice.rb
```diff
@@ -14,23 +14,24 @@ def initialize(fs:, inflector:)
...
 
             fs.mkdir(directory = "slices/#{slice}")
 
             fs.chdir(directory) do
+              fs.write("slice.rb", t("slice.erb", context))
```
timriley, line 23: This actually needs to go into `config/slices/[slice_name].rb` instead of `slices/[slice_name]/slice.rb`.

timriley: And technically, having this file in the first place is completely optional — we don't _need_ the concrete class definition to exist (the framework will define the class for us if it doesn't), but I agree that it probably makes more sense to generate the file than not to. The only downside is that the app may end up feeling a little "heavier" because we're adding files that we don't technically need.

lib/hanami/cli/generators/app/slice.rb
```diff
@@ -14,23 +14,24 @@ def initialize(fs:, inflector:)
             @inflector = inflector
           end
 
-          def call(app, slice, slice_url_prefix, context: SliceContext.new(inflector, app, slice, slice_url_prefix)) # rubocop:disable Metrics/AbcSize
-            fs.inject_line_before_last(fs.join("config", "routes.rb"), /end/, t("routes.erb", context).chomp)
+          def call(app, slice, slice_url_prefix, context: SliceContext.new(inflector, app, slice, slice_url_prefix))
```
timriley, line 17: Just wondering: what's the thinking behind building the SliceContext such that it's injectable as a call arg? Given its just reusing all the previous args, is there any reason we couldn't simplify this method's params signature and just initialize the `SliceContext` inside the method body?

timriley (commented):
@jodosha Looking good! Needs at least one tweak (the location of the slice.rb), and I've left a couple other q's for you :)

lib/hanami/cli/generators/app/slice.rb
jodosha: @timriley If you don't specify the option, it generates a prefix that is the underscored version of the slice name.

lib/hanami/cli/generators/app/slice.rb
jodosha: @timriley The gem needs to be cleaned up. If this isn't a blocker for you, I would not take care of this now.

lib/hanami/cli/generators/app/slice.rb
jodosha: @timriley If not strictly needed, I'd be conservative here and NOT generate for now.

We can always add it to the gen files, but it's harder to remove features.

timriley (approved):
Good one!

PR #33, opened 2022-08-03
Introduce `hanami generate action`

jodosha:
# Feature

Add `hanami generate action` CLI command to Hanami applications.

## Generate Action for App

It performs the following actions:

  * Edit `config/routes.rb` to add a the route
  * Create `app/actions/[controller]` directory
  * Create `app/actions/[controller]/[action].rb` file (with `MyApp::Actions::Books::Index < MyApp::Action` class definition)

Example:

```shell
⚡ bundle exec hanami generate action books.index
```

```shell
⚡ tree app
app
├── action.rb
└── actions
    └── books
        └── index.rb

2 directories, 2 files
```

`config/routes.rb`

```ruby
# frozen_string_literal: true

require "hanami/routes"

module Bookshelf
  class Routes < Hanami::Routes
    define do
      root { "Hello from Hanami" }
      get "/books", to: "books.index"
    end
  end
end
``` 

`app/actions/books/index.rb`

```ruby
# frozen_string_literal: true

module Bookshelf
  module Actions
    module Books
      class Index < Bookshelf::Action
        def handle(_req, res)
          res.body = self.class.name
        end
      end
    end
  end
end
```

## Generate Action for Slice

It performs the following actions:

  * Edit `config/routes.rb` to add a the route within the corresponding `slice` block
  * Create `slices/[slice]/actions/[controller]` directory
  * Create `slices/[slice]/actions/[controller]/[action].rb` file (with `MySlice::Actions::Users::Index < MySlice::Action` class definition)

Example:

```shell
⚡ bundle exec hanami generate action users.index --slice=admin
```

```shell
⚡ tree slices/admin
slices/admin
├── action.rb
└── actions
    └── users
        └── index.rb

2 directories, 2 files
``` 

`config/routes.rb`

```ruby
# frozen_string_literal: true

require "hanami/routes"

module Bookshelf
  class Routes < Hanami::Routes
    define do
      root { "Hello from Hanami" }

      slice :admin, at: "/admin" do
        get "/users", to: "users.index"
      end
    end
  end
end
```

`slices/admin/actions/users/index.rb`

```ruby
# frozen_string_literal: true

module Admin
  module Actions
    module Users
      class Index < Admin::Action
        def handle(*, res)
          res.body = self.class.name
        end
      end
    end
  end
end
```

## Options

```shell
⚡ bundle exec hanami g action --help
Command:
  hanami g action

Usage:
  hanami g action NAME

Arguments:
  NAME                              # REQUIRED Action name

Options:
  --url=VALUE                       # Action URL
  --http=VALUE                      # Action HTTP method
  --slice=VALUE                     # Slice name
  --help, -h                        # Print this help
```

---

Note for the reviewers: specs will be generated by `hanami-rspec`

### DIFF
--- a/lib/hanami/cli/commands/app/generate/action.rb
+++ b/lib/hanami/cli/commands/app/generate/action.rb
@@ -20,13 +20,13 @@ module Hanami
             DEFAULT_SKIP_VIEW = false
             private_constant :DEFAULT_SKIP_VIEW
 
-            argument :slice, required: true, desc: "Slice name"
             argument :name, required: true, desc: "Action name"
             option :url, required: false, type: :string, desc: "Action URL"
             option :http, required: false, type: :string, desc: "Action HTTP method"
             option :format, required: false, type: :string, default: DEFAULT_FORMAT, desc: "Template format"
             option :skip_view, required: false, type: :boolean, default: DEFAULT_SKIP_VIEW,
                                desc: "Skip view and template generation"
+            option :slice, required: false, desc: "Slice name"
 
             def initialize(fs: Dry::Files.new, inflector: Dry::Inflector.new,
                            generator: Generators::App::Action.new(fs: fs, inflector: inflector), **)
@@ -35,8 +35,8 @@ module Hanami
             end
 
             # rubocop:disable Metrics/ParameterLists
-            def call(slice:, name:, url: nil, http: nil, format: DEFAULT_FORMAT, skip_view: DEFAULT_SKIP_VIEW, **)
-              slice = inflector.underscore(Shellwords.shellescape(slice))
+            def call(name:, url: nil, http: nil, format: DEFAULT_FORMAT, skip_view: DEFAULT_SKIP_VIEW, slice: nil, **)
+              slice = inflector.underscore(Shellwords.shellescape(slice)) unless slice.nil?
               name = inflector.underscore(Shellwords.shellescape(name))
               *controller, action = name.split(ACTION_SEPARATOR)
 
@@ -44,7 +44,7 @@ module Hanami
                 raise ArgumentError.new("cannot parse controller and action name: `#{name}'\n\texample: users.show")
               end
 
-              generator.call(slice, controller, action, url, http, format, skip_view)
+              generator.call(controller, action, url, http, format, skip_view, slice)
             end
             # rubocop:enable Metrics/ParameterLists
 
--- a/lib/hanami/cli/commands/app/generate/action.rb
+++ b/lib/hanami/cli/commands/app/generate/action.rb
@@ -36,7 +36,7 @@ module Hanami
 
             # rubocop:disable Metrics/ParameterLists
             def call(name:, url: nil, http: nil, format: DEFAULT_FORMAT, skip_view: DEFAULT_SKIP_VIEW, slice: nil, **)
-              slice = inflector.underscore(Shellwords.shellescape(slice)) unless slice.nil?
+              slice = inflector.underscore(Shellwords.shellescape(slice)) if slice
               name = inflector.underscore(Shellwords.shellescape(name))
               *controller, action = name.split(ACTION_SEPARATOR)
 
--- a/lib/hanami/cli/commands/app/generate/action.rb
+++ b/lib/hanami/cli/commands/app/generate/action.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "hanami/cli/command"
+require "hanami/cli/commands/app/command"
 require "hanami/cli/generators/app/action"
 require "dry/inflector"
 require "dry/files"
@@ -11,7 +11,7 @@ module Hanami
     module Commands
       module App
         module Generate
-          class Action < Command
+          class Action < App::Command
             # TODO: ideally the default format should lookup
             #       slice configuration (Action's `default_response_format`)
             DEFAULT_FORMAT = "html"
@@ -44,7 +44,7 @@ module Hanami
                 raise ArgumentError.new("cannot parse controller and action name: `#{name}'\n\texample: users.show")
               end
 
-              generator.call(controller, action, url, http, format, skip_view, slice)
+              generator.call(app.namespace, controller, action, url, http, format, skip_view, slice)
             end
             # rubocop:enable Metrics/ParameterLists
 
--- a/lib/hanami/cli/commands/app/generate/action.rb
+++ b/lib/hanami/cli/commands/app/generate/action.rb
@@ -23,9 +23,9 @@ module Hanami
             argument :name, required: true, desc: "Action name"
             option :url, required: false, type: :string, desc: "Action URL"
             option :http, required: false, type: :string, desc: "Action HTTP method"
-            option :format, required: false, type: :string, default: DEFAULT_FORMAT, desc: "Template format"
-            option :skip_view, required: false, type: :boolean, default: DEFAULT_SKIP_VIEW,
-                               desc: "Skip view and template generation"
+            # option :format, required: false, type: :string, default: DEFAULT_FORMAT, desc: "Template format"
+            # option :skip_view, required: false, type: :boolean, default: DEFAULT_SKIP_VIEW,
+            #                    desc: "Skip view and template generation"
             option :slice, required: false, desc: "Slice name"
 
             def initialize(fs: Dry::Files.new, inflector: Dry::Inflector.new,
--- a/lib/hanami/cli/generators/app/action/action.erb
+++ b/lib/hanami/cli/generators/app/action/action.erb
@@ -1,4 +1,3 @@
-# auto_register: false
 # frozen_string_literal: true
 
 require "<%= underscored_slice_name %>/action"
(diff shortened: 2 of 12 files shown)


spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -1,299 +1,460 @@
...
+
+    xit "can skip view creation" do
+      within_application_directory do
+        subject.call(name: action_name, skip_view: true)
+
+        expect(fs.exist?("app/actions/#{controller}/#{action}.rb")).to be(true)
+
+        expect(fs.exist?("app/views/#{controller}/#{action}.rb")).to be(false)
+        expect(fs.exist?("app/templates/#{controller}/#{action}.html.erb")).to be(false)
+      end
```
solnic, line 168: Maybe these should be set as pending rather than be skipped?

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -1,299 +1,460 @@
...
+        # expect(fs.read("slices/#{slice}/views/#{controller}/#{action}.rb")).to eq(view_file)
+
+        # template
+        # expect(fs.directory?("slices/#{slice}/templates/#{controller}")).to be(true)
+        #
+        # template_file = <<~EXPECTED
+        #   <h1>#{inflector.classify(slice)}::Views::#{inflector.camelize(controller)}::#{inflector.classify(action)}</h1>
+        #   <h2>slices/#{slice}/templates/#{controller}/#{action}.html.erb</h2>
+        # EXPECTED
+        # expect(fs.read("slices/#{slice}/templates/#{controller}/#{action}.html.erb")).to eq(template_file)
```
solnic, line 252: Is it really needed to leave this as commented out vs just removing it?

lib/hanami/cli/commands/app/generate/action.rb
```diff
@@ -35,16 +35,16 @@ def initialize(fs: Dry::Files.new, inflector: Dry::Inflector.new,
             end
 
             # rubocop:disable Metrics/ParameterLists
-            def call(slice:, name:, url: nil, http: nil, format: DEFAULT_FORMAT, skip_view: DEFAULT_SKIP_VIEW, **)
-              slice = inflector.underscore(Shellwords.shellescape(slice))
+            def call(name:, url: nil, http: nil, format: DEFAULT_FORMAT, skip_view: DEFAULT_SKIP_VIEW, slice: nil, **)
+              slice = inflector.underscore(Shellwords.shellescape(slice)) unless slice.nil?
```
solnic, line 39:
```suggestion
              slice = inflector.underscore(Shellwords.shellescape(slice)) if slice
```

solnic (approved):
👏🏻 I only left some minor questions/suggestion

lib/hanami/cli/commands/app/generate/action.rb
```diff
@@ -35,16 +35,16 @@ def initialize(fs: Dry::Files.new, inflector: Dry::Inflector.new,
...
               name = inflector.underscore(Shellwords.shellescape(name))
               *controller, action = name.split(ACTION_SEPARATOR)
 
               if controller.empty?
                 raise ArgumentError.new("cannot parse controller and action name: `#{name}'\n\texample: users.show")
               end
 
-              generator.call(slice, controller, action, url, http, format, skip_view)
+              generator.call(app.namespace, controller, action, url, http, format, skip_view, slice)
```
timriley, line 47: I wonder whether we really need the "controller" terminology here at all. I know this is mostly internal code but AFAICT we'll only ever refer to these as plain "actions" everywhere else.

This is really just an (optionally nested) action path, e.g. "foo", or "foo.bar.baz". Is there any reason we couldn't just pass that through as a singular string and then work out the nested directory aspect over in the generator?

IMO this would make the generator easier to understand, because I didn't really know what the "controller" arg meant when I was just looking at it now. I had to flip back up to here to figure it out.

(I'd be happy for this to be a follow up refactor, in the spirit of not wanting to block this PR!)

lib/hanami/cli/generators/app/action/action.erb
```diff
@@ -1,12 +1,12 @@
...
 # frozen_string_literal: true
 
-require "<%= underscored_slice_name %>/action"
-
-module <%= classified_slice_name %>
+module <%= classified_app_name %>
   module Actions
 <%= module_controller_declaration %>
-<%= module_controller_offset %>class <%= classified_action_name %> < <%= classified_slice_name %>::Action
+<%= module_controller_offset %>class <%= classified_action_name %> < <%= classified_app_name %>::Action
```
timriley, line 6: (Just a comment, not a request for change!)

That we have to use this `module_controller_offset` as our "spacer.gif" here is kinda hilarious. I've always wished when writing generators that we could just spit out whatever gross looking Ruby and then have it run through a formatter before being saved to disk. Maybe one day!

lib/hanami/cli/generators/app/slice/action.erb
```diff
@@ -1,8 +1,6 @@
 # auto_register: false
 # frozen_string_literal: true
 
-require "<%= underscored_app_name %>/action"
-
```
timriley, line 5: Love to remove a require!

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -1,299 +1,460 @@
...
-    expect(fs.read("config/routes.rb")).to match(%(get "/users", to: "users.index"))
+          require "hanami/routes"
 
-    # action
-    expect(fs.directory?(directory = "slices/#{slice}/actions/#{controller}")).to be(true)
+          module #{app}
+            class Routes < Hanami::Routes
+              define do
+                root { "Hello from Hanami" }
+                get "/users", to: "users.index"
```
timriley, line 34: Just out of interest, is the leading slash in `"/users"` here actually necessary? I might be remembering things wrong, but I recall mostly not using it...

timriley (approved):
Looks good!

I have one big question that I think I'd like to see us change (or at least us discuss), which is the change to generating `def handle(req, res)`.

The rest are just minor thoughts/questions and not blockers :)

lib/hanami/cli/commands/app/generate/action.rb
jodosha: Let's cleanup later then. I'm going on vacation in less than 3hrs 😎 🌴 .

lib/hanami/cli/generators/app/action/action.erb
```diff
@@ -1,12 +1,12 @@
...
+<%= module_controller_offset %>  def handle(*, res)
```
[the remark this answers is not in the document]
jodosha: @timriley Not binding `req`/`res` when not needed it's faster.

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
jodosha: @solnic When we'll have views (hopefully soon), it will be a matter of uncommenting those lines.

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
jodosha: @timriley IDK if that would work, maybe accidentally.

But why would you declare a route w/o the leading slash?
That is a Relative URL, which has the leading slash.

timriley: I must just be mis-remembering, all good!

lib/hanami/cli/generators/app/action/action.erb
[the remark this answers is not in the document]
timriley: @jodosha Is it really _meaningfully_ faster? Are we willing to do that at the cost of discoverability here?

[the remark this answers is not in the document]
cllns: I agree about including both request and response in the default.

Can some Rubocop rule can automatically fix things like that?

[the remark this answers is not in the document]
jodosha: @timriley +100 req/s.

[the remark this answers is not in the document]
cllns: Additionally, I'd really like to see them spelled out as `def handle(request, response)` but I guess that is unpopular.

`req` and `res` are only one character apart, so they're easy to mess up. And, while *we're* used to it now, `res` especially isn't a common abbreviation for `response`. In context, sure, one can figure it out but it's not as obvious.

From a documentation perspective, I was thinking we could consistently use `request` and `response` for clarity, as well. But we should be consistent so users don't get misnamed variable errors.

(I've held onto this feedback for a year or so because it felt irrelevant to the end-user experience but now it's very relevant with a generator 😅 )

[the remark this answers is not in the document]
timriley: To give us something we can release this week, I'm inclined to change this to:

```
def handle(*, response)
```

as a starting point.

I'm definitely convinced by the value of using the full words here (which is _much_ more common than the `req`/`res` abbreviations used here so far).

We can then have a separate discussion about the discoverability of the first `request` parameter versus any performance improvements gained from splatting it out.

timriley:
Have changed the action generator to create `def handle (*, response)`. Will merge this now so we can include it in today's release!

PR #42, opened 2022-09-03
Add required hanami-controller and hanami-view to the Gemfile

waiting-for-dev:
Since hanami/hanami#1204, we're no longer returning a null configuration
when action/view/router are not present. We need to add all the required
gems to the Gemfile to test their related commands.


waiting-for-dev:
The PR on hanami has been reverted, so closing it for now.

PR #44, opened 2022-09-07
fix: Use app[:settings] in database commands.

pat:
This is so the existing database commands can continue to work (albeit unofficially) with hanami 2.0.0.beta2.

Yes, I realise they're not properly supported with the beta releases, but they have still worked when using a custom `bin/hanami` shim that re-enables them. So, this is just a small change to help that remain the case. 😅

### DIFF
--- a/lib/hanami/cli/commands/db/utils/database.rb
+++ b/lib/hanami/cli/commands/db/utils/database.rb
@@ -30,7 +30,7 @@ module Hanami
             }.freeze
 
             def self.[](app)
-              config = DatabaseConfig.new(app.settings.database_url)
+              config = DatabaseConfig.new(app[:settings].database_url)
 
               resolver = SCHEME_MAP.fetch(config.db_type) do
                 raise "#{config.db_type} is not a supported db scheme"


timriley:
Hey Pat! I actually have an open branch from just a few days ago where I’ve done similar, except I changed it to use DATABASE_URL from the ENV directly, with the idea being that it would be nicer if we didn’t have to have “framework-level” concerns eventually go into every user’s settings.rb.

This is of course still subject to change when we do get to the point of polishing the db integration, but would it work for you in the meantime?

pat:
Sounds wise with the separation of concerns, and it would work fine in our setup. I do wonder - would it be worth having a known set of variables alongside the app's own settings (i.e. a wrapper around specific ENV values), instead of just using `ENV.fetch` everywhere?

timriley:
@pat This is resolved by #50 now, which I've just merged. FYI I updated it to check `app[:settings].database_url` first, and fall back to `ENV` if that setting is not defined.

This may not be the long term behavior, but it feels like a reasonable bridging approach for now.

timriley:
Resolved by #50.

pat:
Thanks Tim, all sounds good to me.

PR #46, opened 2022-09-13
Fix routes definition in test app

cllns:
No longer needs (or can use) the  block

https://github.com/hanami/hanami/pull/1208

Note that this is different from #45, which is about the apps that this cli generates, whereas this is to make the test suite for this repo green again.

### DIFF
--- a/spec/fixtures/test/config/routes.rb
+++ b/spec/fixtures/test/config/routes.rb
@@ -4,9 +4,7 @@ require "hanami/routes"
 
 module Test
   class Routes < Hanami::Routes
-    define do
-      get "/", to: "home.index"
-      get "/about", to: "home.about"
-    end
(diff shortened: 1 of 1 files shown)


solnic:
I'm confused. I started working on the same thing in update-routes-generator branch but the changes required to make the specs pass are much bigger. Are we not running the whole spec suite on CI?

cllns:
It looks like we're running them all. What this PR does is get the test suite to run again (the issue was the fixture app causing `Hanami::SliceLoadError`, not a spec failures) but the code it's generating is invalid now, since it has the define block. Your PR fixes that to make sure it's generating valid apps.

I guess this repo doesn't test that the files being created and modified are still correct, which we likely want to add later.

PR #49, opened 2022-09-18
Add console output when generating files

cllns:
The main feature being added here is outputting to the console when a file is created/updated.

https://trello.com/c/HGvL6mD9/213-add-output-to-cli-commands

Looking for feedback on the approach taken here. Still (quite a lot!) more to do but you can get the general idea here.

A couple differences in output compared to Hanami 1.x and Rails is:
1. use the past-tense ("Created", "Updated"), and only print that _after_ the operation. Not committed to this, it just makes more sense to me.
2. show (the first line of) what's been added when adding to the end of a block. This would be extended to the other similar methods. Not committed to this either, just seems like a nice touch.

Long-term, I'd be interested in seeing this particular pairing of `Dry::CLI` and `Dry::Files` extracted to `Dry::Files::Generator` or `Dry::CLI::Generator` (or even a dry-generator gem), but this concrete use-case should be established and  battle-tested first.


spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -5,17 +5,22 @@
...
-  let(:generator) { Hanami::CLI::Generators::App::Action.new(fs: fs, inflector: inflector) }
+  let(:action_generator) { Hanami::CLI::Generators::App::Action.new(fs: fs, out: stdout, inflector: inflector) }
+  let(:stdout) { StringIO.new }
   let(:app) { Hanami.app.namespace }
   let(:dir) { inflector.underscore(app) }
   let(:controller) { "users" }
   let(:action) { "index" }
   let(:action_name) { "#{controller}.#{action}" }
 
+  def output
```
cllns, line 20: Any better ideas? Can't memoize it as a `let` with how the spec are structured now (particularly the `"infers RESTful action URL and HTTP method for routes"` specs)

jodosha (changes requested):
@cllns Thanks for taking care of this feature, but I'm not convinced about the solution.

Mainly three things:

1. You want to inject a generator (`Hanami::CLI::Generator`) inside a generator (`Hanami::CLI::Generators::App::Action`). This seems conceptually wrong.
2. You're using the generator as a decorator for `dry-files`. I see `Hanami::CLI::Generator` to constantly follow `dry-files` interface changes.
3. You're seeking for an abstraction, before to have this implemented.

---

Would be possible to implement this directly inside `Hanami::CLI::Generators::App::Action`? 

We already have all the raw dependencies.

It would make the code more procedural, but at least we'll have a good base to refactor on a later stage.

This is the simplest thing that could work.

cllns:
Yea, I realized the trouble with the naming when I started extending it to the slice generator yesterday, which already has a `@generator` instance variable, so there's an obvious naming conflict.

Instead, I decided to try it as `Hanami::CLI::Files`: a subclass of `Dry::Files` that prints the info we want (these changes are pushed here now). Obviously we should be careful with inheritance, but it really does feel like a form a _specialization_ to me, where inheritance can be appropriate. What do you think about this? This seems like less work, just swapping out the `fs` param and getting the printing for "free".

If you still think I should take the procedural approach and add a bunch of `puts`s throughout the codebase, I can take that approach instead. I hear your point about implementing this feature first, it just seemed like a lot of (potentially typo-prone) work that could be avoided by jumping right to the right abstraction. A swing and miss on the first attempt :)

solnic:
I'm OK with inheritance, the code changes look reasonable to me. FWIW, having dry-files a notification system would be best so that you could listen to its events and do whatever you want. For now though, inheritance seems like a simpler choice to get what we want, and we definitely want to see output when running hanami generators 😅

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
solnic: This is fine

solnic (approved):
I'm accepting this already. This feature really seems important and personally I'd like to see this done, excuse the word, ASAP 🙂

timriley (approved):
I agree with @solnic that a notifications system for dry-files would be the ideal way to handle this, but absent that (and we should add it in the future!), the subclass approach here seems like a practical way to achieve what we want with minimal overall disruption, which is important given our release timing. We can iterate on the design later :)

So a tick from me. Let's get this in. What's left to do here, @cllns?

jodosha (approved):
@cllns This looks much better now. Thanks for taking the time!

cllns:
Updated the rest! Let me know your thoughts.

A few things: 
1. Using `chdir` messes up the output, because it's relative to the new location, rather than from the beginning. 
2. Similarly, `hanami new bookshelf` outputs `Created bookshelf/` then the rest are relative to within that folder. I'm not sure what's preferable here: listing all the new files with that as a prefix, or just creating a way to override the output of that altogether. It does seems like important info to surface,
3. We have some duplicated variables now, `fs` being Dry::Files in `Hanami::CLI::Command`. I don't really understand the current structure entirely. Similarly, that command has an `out` instance variable that we are _not_ using, since we inject it in our classes too, and I'm not sure how it was intended to work.

### `hanami new`
```sh
% hanami new bookshelf
Created bookshelf/
Created .env
Created README.md
Created Gemfile
Created Rakefile
Created config.ru
Created config/app.rb
Created config/settings.rb
Created config/routes.rb
Created config/puma.rb
Created lib/tasks/.keep
Created lib/bookshelf/types.rb
Created app/actions/.keep
Created app/action.rb
```
### `hanami generate action`
```sh
% hanami generate action home.index
Updated config/routes.rb
Created app/actions/home/
Created app/actions/home/index.rb
```

### `hanami generate slice`
```sh
% hanami generate slice admin
Updated config/routes.rb
Created slices/admin/
Created slices/admin/action.rb
Created slices/admin/actions/.keep
```

### `hanami generate action --slice`
```sh
% hanami generate action --slice admin home.index
Updated config/routes.rb
Updated slices/admin/actions/home/index.rb
```

### `hanami generate action --slice` **NESTED** name
```sh
% hanami generate action --slice admin books.bestsellers.nonfiction.index
Updated config/routes.rb
Created slices/admin/actions/books/bestsellers/nonfiction/
Created slices/admin/actions/books/bestsellers/nonfiction/index.rb
```

solnic:
Please rebase, I've addressed spec failures in main.

cllns:
Just added another commit and updated the previous comment (including output) to reflect it. (I added some output for `fs.chdir`)

solnic (approved):
🚀 👍🏻

solnic:
@cllns could we get this merged in?

timriley:
Let's do it. If it works like we want, we want this merged before the beta4 release that'll go out on Monday.

timriley:
Thank you for your work here, @cllns!

PR #50, opened 2022-09-20
Make database setup compatible with latest betas (and fall back to ENV if required)

timriley:


jodosha (approved):

solnic:
Why latest betas don't guarantee `settings[:database_url]` to be present?

timriley:
If config/settings.rb is absent, then we don't register the settings component at all.

See https://github.com/hanami/hanami/blob/b1e0930241133435a0f90625d405434d08b5c9e0/lib/hanami/providers/settings.rb#L25 for the code. 

This is the same behavior as we have for routing-related components too.

timriley:
Oh wait, you were asking about the database_url setting itself. 

It'd because that setting right now just be _user-defined_. We don't generate it out into the settings file, so it won't always be there. 

I'm not sure we should be requiring every app to have a `setting :database_url` line in their settings files either. This feels more like framework setup than app-specific setup, if you know what I mean. 

So this is why I provided the ENV fallback for now. I'm using it inside timriley/decafsucks.

solnic:
Oh I see, thanks!

PR #61, opened 2022-11-13
Update API docs for 2.0.0

timriley:


timriley:
[Prepare API docs for hanami-cli](https://trello.com/c/RLCsK9bk/269-prepare-api-docs-for-hanami-cli)

lib/hanami/cli/generators/app/action_context.rb
```diff
@@ -6,50 +6,72 @@
 module Hanami
   module CLI
     module Generators
+      # @since 2.0.0
+      # @api private
       module App
+        # @since 2.0.0
+        # @api private
```
timriley, line 13: @jodosha I marked this as `@api private` (with a view to having only the minimum required public API for this gem) but later noticed that these are also used by hanami-rspec.

Do you think they should be public API? Or are we happy just keeping these as internal for now?

lib/hanami/cli/commands/app/install.rb
```diff
@@ -6,6 +6,19 @@ module Hanami
...
+        # The `install` command exists to provide third parties a hook for their own installation
+        # behaviour to be run as part of `hanami new`.
+        #
+        # Third parties should register their install commands like so:
+        #
+        # ```ruby
+        # if Hanami::CLI.within_hanami_app?
+        #   Hanami::CLI.after "install", MyHanamiGem::CLI::Commands::Install
+        # end
+        # ````
```
katafrakt, line 18: Screenshot 2022-11-14 at 14 01 40

This doesn't render well, even though the markup for YARD is set to Markdown. I think it silently ignores this setting if it cannot find any markdown gem present. I added redcarpet to the `Gemfile` and after that it starts to render correctly:

Screenshot 2022-11-14 at 14 11 40

lib/hanami/cli/commands/app/middleware.rb
```diff
@@ -23,10 +23,16 @@ module App
         # $ bundle exec hanami middleware --with-arguments
         # /    Rack::Session::Cookie args: [{:secret=>"foo"}]
         # ```
+        #
```
katafrakt, line 26: Oddly, the above code blocks render fine without redcarpet, so it's probably because of language being set.

lib/hanami/cli/generators/app/action_context.rb
jodosha: @timriley Mark as private.
Even if it's `hanami-rspec`, it's private for us as Hanami team.
I don't expect other people to use this class.

PR #62, opened 2022-11-13
Convert to Zeitwerk for code loading

timriley:
Covert the gem to use Zeitwerk for code loading. To do this:

- Add zeitwerk to the gemspec
- Add `lib/hanami-cli.rb` so the gem follows a conventional structure
- Set up the Zeitwerk loader inside `lib/hanami/cli.rb` (call it `.gem_loader` instead of `.loader` like we have done in other gems, since `Hanami::CLI` is a class with other responsibilities, and `loader` is too generic a term)
- Remove all now-unneeded manual requires for files inside the gem

This is not a breaking change because any existing `require "hanami/cli"` (which is one one and only entrypoint for this gem) will continue to work as expected, since that is where the Zeitwerk loader setup is done.


timriley:
[Zeitwerkify hanami-cli](https://trello.com/c/WImuE9lL/331-zeitwerkify-hanami-cli)

PR #63, opened 2022-11-15
In case of exception, print error message and exit

jodosha:
Don't print stack trace to stderr.

---

Manual tested:

```shell
⚡ bundle exec hanami generate action users.index --slice=foo
slice `foo' is missing, please generate with `hanami generate slice foo'
⚡ echo $?
1
```

---

https://trello.com/c/rtZynx28

### DIFF
--- a/exe/hanami
+++ b/exe/hanami
@@ -7,4 +7,10 @@ require "hanami/cli"
 cli = Dry::CLI.new(Hanami::CLI)
 Hanami::CLI::Bundler.require(:cli)
 
-cli.call
+begin
+  cli.call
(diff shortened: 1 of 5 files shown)


timriley (approved):

exe/hanami
```diff
@@ -7,4 +7,9 @@ require "hanami/cli"
 cli = Dry::CLI.new(Hanami::CLI)
 Hanami::CLI::Bundler.require(:cli)
 
-cli.call
+begin
```
jodosha, line 10: @timriley @solnic I believe that this should be a `dry-cli` feature.

PR #65, opened 2022-11-18
Use Hanami.app_path to find if we're in an app

timriley:
This allows the hanami CLI to expose its app commands even if invoked from a deeper app directory.

Since this may have caused an issue with the CLI properly executing in earlier testing (before our 2.0.0.beta3 release), wrap the call to `Hanami.app_path` in a rescue that will fall back to a simpler mechanism that will still work in the common cases.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,6 +16,8 @@ Naming/HeredocDelimiterNaming:
 Naming/MethodParameterName:
   AllowedNames:
     - fs
+Naming/RescuedExceptionsVariableName:
+  Enabled: false
 Style/AccessorGrouping:
   Enabled: false
 Style/BlockDelimiters:
--- a/lib/hanami/cli/commands.rb
+++ b/lib/hanami/cli/commands.rb
@@ -12,8 +12,15 @@ module Hanami
     # @api private
     # @since 2.0.0
     def self.within_hanami_app?
-      File.exist?("config/app.rb") ||
-        File.exist?("app.rb")
+      require "hanami"
+
+      !!Hanami.app_path
+    rescue LoadError => e
+      raise e unless e.path == "hanami"
+
+      # If for any reason the hanami gem isn't installed, make a simple best effort to determine
+      # whether we're inside an app.
+      File.exist?("config/app.rb") || File.exist?("app.rb")
     end
 
     # Contains the commands available for the current `hanami` CLI execution, depending on whether


timriley:
[Restore CLI working in deeper app directories](https://trello.com/c/LrtaVkNq/327-restore-cli-working-in-deeper-app-directories)

timriley:
@jodosha we need Hanami.app path available to hanami so that `require "hanami/setup"` works from any nested dir, much like what we want to achieve here for the CLI; see `Hanami.setup`:

```ruby
  def self.setup(raise_exception: true)
    return app if app?

    app_path = self.app_path

    if app_path
      prepare_load_path
      require(app_path.to_s)
      app
    elsif raise_exception
      raise(
        AppLoadError,
        "Could not locate your Hanami app file.\n\n" \
        "Your app file should be at `config/app.rb` in your project's root directory."
      )
    end
  end
```

I've manually tested this change to hanami-cli and it works as expected.

Before this change, if you `cd lib/` inside an Hanami app, you got this:

```
$ cd lib
$ bundle exec hanami
Commands:
  hanami new APP         # Generate a new Hanami app
  hanami version         # Hanami version
```

With this change, you instead get this:

```
# cd lib
$ bundle exec hanami
Commands:
  hanami console                              # Start app console (REPL)
  hanami generate [SUBCOMMAND]
  hanami install                              # Install Hanami third-party plugins
  hanami middleware                           # Print app Rack middleware stack
  hanami routes                               # Print app routes
  hanami server                               # Start Hanami app server
  hanami version                              # Print Hanami app version
```

The fact of the matter is, if you _are_ inside an Hanami app, then the `hanami` gem is going to be available, right?

And we actually require hanami elsewhere inside hanami-cli, see `Hanami::CLI::Commands::App::Command.app` for example:

```ruby
          def app
            @app ||=
              begin
                require "hanami/prepare"
                Hanami.app
              end
          end
```

That's pretty much the same pattern as we're following for this change, right? Require files from the hanami gem, but be prepared to bail out gracefully if they're not available.

You touch on this being not a gemspec dependency, but we also have many such optional dependency cases already (e.g. hanami itself on the other various gems), and we seem to be handling them responsibly enough.

Given this is needed inside hanami _and_ hanami-cli, the only other feasible option would be to move this to hanami-utils, but I don't feel that's the best outcome. Since finding the app_path is critical to the boot process of hanami, I'd feel more confident in this method staying in the hanami gem right next to the code that critically depends on it.

Do any of these points change your perspective?

jodosha (approved):
@timriley Alright, let's move on with this. We can rearrange private code later on. :shipit:

timriley:
Thanks @jodosha, definitely happy to revisit this later on! And in the meantime, very happy to have this ticked off and working nicely again :)

Issue #59, opened 2022-11-08
No warning for `hanami new` when the directory already exists


janko:
I ran `hanami new blog`, but didn't realize the current working directory already had a `blog/` directory, where I had previously generated a Rails application. The generator happily merged Hanami files into the Rails codebase. It would be great if the `hanami new` command would abort if the target directory already exists 🙂

janko:
Sorry, this appears to be a duplicate of https://github.com/hanami/cli/issues/52 😛

jodosha:
@janko Fixed in https://github.com/hanami/cli/pull/66

PR #66, opened 2022-11-18
Raise exception if trying to create Hanami app in existing folder (or file)

### LINKED ISSUE #52: Add warning/confirmation when creating a new project in a folder that already exists
Right now, `hanami new` creates the files and folder regardless of whether any exist.

Instead of doing that, we should prompt the user to ask if it's OK to add files to their existing folder, or to overwrite an existing file. Or, we could just refuse to create a project in an existing folder, to avoid dealing with overwriting issues.

```shell
% mkdir foo
% echo "HELLO_WORLD" > foo/README.md
% touch foo/TEST_FILE
% ls foo/
TEST_FILE
% hanami new foo
% ls foo
Gemfile       Gemfile.lock  Guardfile     README.md     Rakefile      TEST_FILE app           config        config.ru     lib           spec
% cat foo/README.md
# Foo
```

`foo/README.md` demonstrates that files are overwritten and `TEST_FILE` shows that existing files aren't deleted.

This same issue and approach should be taken for generating files within a project, e.g. `hanami generator action users.show` if the users.show action already exists. We don't want to accidentally overwrite the user's file. That can optionally be handled separately, but we should make an issue for that if it doesn't get resolved with the PR that closes this issue.

cllns:
Closes #52 (and reported in #59 by someone who is not me)

```bash
% mkdir bookshelf # or `touch bookshelf`
% exe/hanami new bookshelf
Cannot create new Hanami app in an existing folder bookshelf/
```

This is simpler than prompting the user to decide if we should still proceed. They can decide or move folders around.

### DIFF
--- a/lib/hanami/cli/errors.rb
+++ b/lib/hanami/cli/errors.rb
@@ -22,7 +22,7 @@ module Hanami
 
     class PathAlreadyExistsError < Error
       def initialize(path)
-        super("Cannot create new Hanami app in an existing file or folder: #{path}")
+        super("Cannot create new Hanami app in an existing path: `#{path}'")
       end
     end
 
--- a/lib/hanami/cli/errors.rb
+++ b/lib/hanami/cli/errors.rb
@@ -20,6 +20,12 @@ module Hanami
       end
     end
 
+    class PathAlreadyExistsError < Error
+      def initialize(path)
+        super("Cannot create new Hanami app in an existing file or folder: #{path}")
+      end
+    end
+
     class MissingSliceError < Error
       def initialize(slice)
         super("slice `#{slice}' is missing, please generate with `hanami generate slice #{slice}'")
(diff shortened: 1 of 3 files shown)


jodosha:
@cllns I don’t want to rush this change in 2.0.
Because the UX isn’t clear yet and I would extend this to all hanami commands that touch file system.
Better to tackle the problem organically.

jodosha (commented):
@cllns I don’t want to rush this change in 2.0.
Because the UX isn’t clear yet and I would extend this to all `hanami` commands that touch file system.
Better to tackle the problem organically.

jodosha (approved):

lib/hanami/cli/errors.rb
```diff
@@ -20,6 +20,12 @@ def initialize(message)
       end
     end
 
+    class PathAlreadyExistsError < Error
+      def initialize(path)
+        super("Cannot create new Hanami app in an existing file or folder: #{path}")
```
jodosha, line 25:
```suggestion
        super("Cannot create new Hanami app in an existing path: `#{path}'")
```

PR #70, opened 2022-12-01
Load dotenv and use hanami port for hanami server

jodosha:
Fixes https://github.com/hanami/hanami/issues/1264
Ref https://github.com/hanami/hanami/pull/1266


jodosha:
Confirmed by @kucho https://github.com/hanami/hanami/issues/1264#issuecomment-1334445899

PR #71, opened 2022-12-07
Fix 'which' for windows

dsisnero:
which("bundler") on windows fails to find bundler because on windows the cmd has a ".bat" extension

This commit changes which(cmd) to loop throught PATHEXTS if they exist to find the correct cmd on windows.


jodosha (changes requested):
@dsisnero Thanks for this PR. Can you please make the patch less invasive? For us it's important to preserve Git history.

This patch rewrites the whole file, but the actual changes seems to be only for `#which` method.

PR #72, opened 2022-12-08
Fix which function for windows

dsisnero:
Which on windows doesn't find the cmd becuase they have extensions [.bat, .cmd] .  This enables which to find cmds on windows


jodosha (approved):
@dsisnero Thanks for this enhancement! 👍

Issue #73, opened 2022-12-18
Absence of `.gitignore` pollutes application git history


vikdotdev:
Recently finally had the time to play properly with new version of Hanami and encountered an annoyance related to [this quote](https://github.com/hanami/cli/pull/47#issue-1373368724). I did the usual flow when making a new project and uploading it to Github:
```
hanami new bookshelf
cd bookshelf
# here I followed the Hanami guide, added and ran specs etc.

git init
git add .
git ci -m "Initial commit"
# then set up origin and push to remote
```
Given that `hanami/cli` does not include any git related files e.i `.gitignore` I ended up accidentally commiting a 1k+ `log/test.log` file which pollutes the history and is simply annoying.

I imagine a lot of people follow similar steps of simply copy commands from Github and trusting this won't happen. Adding a `.gitignore` to `hanami new` command can fix this. I'm can help with a contribution if you think this is something worth adding.

solnic:
I agree. This should be added automatically as it's a pragmatic thing to do.

jodosha:
@vikdotdev Can you please have a look at #77?

PR #77, opened 2023-01-26
`hanami new` to generate a default `.gitignore`

jodosha:
Closes https://github.com/hanami/cli/issues/73


jodosha:
Got @vikdotdev approval, which is the reporter of #73

PR #80, opened 2023-02-28
Fix rollback command

charithperera:
## Purpose

Fix rollback command as its currently not working correctly.

## Context

* The current implementation of rollback is passing the incorrect target to the `database` class. It's passing the target version of the latest migration instead of the previous one. This is resulting in no rollback occurring. 
* Another issue is that the block being passed to `measure` does not return true if successful, resulting in a failure message displaying all the time, even on successful rollbacks.

## Change

* Make the rollback command a bit smarter by allowing it to figure out correctly what the applied migrations are and what the intention is based on if a target is provided or not.
* Pass the `latest_migration - 1` in to ensure the rollback target is one prior to the latest applied migration
* Wrap ` Sequel::TimestampMigrator.new(migrator.connection, migrations_path, {})` in `ROM::SQL.with_gateway(gateway)` to prevent reoccurring error

### DIFF
--- a/lib/hanami/cli/commands/app/db/rollback.rb
+++ b/lib/hanami/cli/commands/app/db/rollback.rb
@@ -32,6 +32,8 @@ module Hanami
 
               measure "database #{database.name} rolled back to #{migration_name}" do
                 database.run_migrations(target: Integer(migration_code))
+
+                true
               end
 
               run_command Structure::Dump if dump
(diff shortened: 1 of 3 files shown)


timriley:
Thanks for this (and hi! 👋🏼), @charithperera. I should be able to give this a proper look early next week :)

timriley (approved):
Really solid work here @charithperera! Well structured, commented, and explained. Tested it out, works as advertised. No other notes! In it goes! 🚀 Thank you for contributing to Hanami! ❤️

PR #81, opened 2023-04-04
Update psql drop_command to force dropdb

Andsbf:
**Context:**

The reason behind it is to improve developer experience, as right now when running a command like `hanami db reset` it will fail if the user has any other ongoing connection, like a connection to db client(like dbeaver), or a `hanami console` session.

image

Couple other possible approaches are:

```ruby
              # create a dedicated method for force drop
              def force_drop_command
                system(cli_env_vars, "dropdb --force #{escaped_name}")
              end
```
```ruby
              # force as an option/param
              def drop_command(force=false)
              command_prefix =  force ? "dropdb --force" : "dropdb"
              
                system(cli_env_vars, "#{command_prefix} #{escaped_name}")
              end
```

**Reference:** 
https://www.postgresql.org/docs/current/sql-dropdatabase.html#id-1.9.3.108.6

### DIFF
--- a/lib/hanami/cli/commands/app/db/drop.rb
+++ b/lib/hanami/cli/commands/app/db/drop.rb
@@ -13,7 +13,7 @@ module Hanami
 
             # @api private
             def call(force: false)
-              if database.drop_command(force)
+              if database.drop_command(force:)
                 out.puts "=> database #{database.name} dropped"
               else
                 out.puts "=> failed to drop #{database.name}"
--- a/lib/hanami/cli/commands/app/db/drop.rb
+++ b/lib/hanami/cli/commands/app/db/drop.rb
@@ -12,8 +12,8 @@ module Hanami
             desc "Delete database"
 
             # @api private
-            def call(**)
-              if database.drop_command
+            def call(force: false)
+              if database.drop_command(force)
                 out.puts "=> database #{database.name} dropped"
               else
                 out.puts "=> failed to drop #{database.name}"
--- a/lib/hanami/cli/commands/app/db/utils/postgres.rb
+++ b/lib/hanami/cli/commands/app/db/utils/postgres.rb
@@ -23,7 +23,7 @@ module Hanami
 
               # @api private
               def drop_command
-                system(cli_env_vars, "dropdb #{escaped_name}")
+                system(cli_env_vars, "dropdb --force #{escaped_name}")
               end
 
               # @api private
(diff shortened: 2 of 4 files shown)


parndt:
I don't think we should use force by default, but making it an option would be 👍

timriley:
Hi @Andsbf! Thanks for this confibution :) Like @parndt suggested, I think this would work best as a `--force` option to the `db drop` command, rather than it being the default.

Would you mind adjusting this PR to turn it into an option? Thanks!

lib/hanami/cli/commands/app/db/utils/postgres.rb
```diff
@@ -22,8 +22,10 @@ def create_command
               end
 
               # @api private
-              def drop_command
-                system(cli_env_vars, "dropdb #{escaped_name}")
+              def drop_command(force: false)
+                command = force ? "dropdb --force" : "dropdb"
```
Andsbf, line 26: I'm not sure what would be your favourite flavour here this or that:
```ruby
                command = "dropdb"

                if force
                  command << " --force"
                end
```

lib/hanami/cli/commands/app/db/utils/sqlite.rb
```diff
@@ -17,7 +17,7 @@ def create_command
               end
 
               # @api private
-              def drop_command
+              def drop_command(**)
```
Andsbf, line 20: seems like "unlink" doesn't support any params, so nothing to do here
https://apidock.com/ruby/Pathname/unlink

Andsbf:
@parndt @timriley  thanks for the follow up, I have updated the PR according to the conversation above, let me know if I got it right. I could not find document to update regard that interface change (making the `Hanami::CLI::Commands::App::DB::Drop`)  take an optional value, so let me know if I miss it.

I would appreciate some help here on how to use it....given I have the following on my `hanami` file:
image

if this PR gets merged I would update my "db drop" register command to this?

```ruby 
Hanami::CLI.register "db drop", Hanami::CLI::Commands::App::DB::Drop.new(force: true)
```

also would the other commands "follow" on the "force" ? I mean, if I call "db reset" would it use the "db drop"  with the `force: true` ? looking at this file I would assume it wouldn't:

https://github.com/hanami/cli/blob/main/lib/hanami/cli/commands/app/db/reset.rb

I highly appreciate you guys putting the time to look at this PR, thanks again!

lib/hanami/cli/commands/app/db/drop.rb
```diff
@@ -12,8 +12,8 @@ class Drop < App::Command
             desc "Delete database"
 
             # @api private
-            def call(**)
-              if database.drop_command
+            def call(force: false)
+              if database.drop_command(force)
```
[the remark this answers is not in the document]
Andsbf: spot on, my bad!

PR #84, opened 2023-05-07
View generators

timriley:
Add support for generating views:

- When generating a new app, via `hanami new`
- When generating an action, via `hanami generate action`
- Independently, via `hanami generate view`


timriley:
[Generate views via CLI commands](https://trello.com/c/FKhgoanp/17-generate-views-via-cli-commands)

PR #88, opened 2023-06-15
Generate public/ error pages in new app

timriley:
I was tempted to copy Rails' CSS for these, but that felt a bit on the nose, so for now, these are non-styled HTML, and there is now the opportunity for someone to come in and help create nicer looking default error pages in the future :)


timriley:
[Generate public/{500,404}.html in new apps](https://trello.com/c/Fj3BH2C9/52-generate-public-500404html-in-new-apps)

PR #53, opened 2022-10-04
Tidy `hanami` help

jodosha:
# Enhancement

Tidy `--help` of all the `hanami` commands.

---

# CHANGELOG

* `hanami middlewares` -> `hanami middleware`
* `hanami generate slice --url-prefix` -> `--url` for parity with `hanami generate action`, which accepts a `--url` option
* `hanami new --skip-bundle` --> `--skip-install` to communicate that both Bundler and Hanami installation are skipped 

 
---

# Code Changes

* `Hanami::CLI::Command#initialize` superclass injects `err:` as dependency. It defaults to `$stderr`.
* `Hanami::CLI::Command::App::Console` renamed REPL to Engine to low level adapter like IRB, Pry, etc..

---

# CLI Output

## Gem commands

Outside an application

### `hanami`

```shell
⚡ hanami --help
Commands:
  hanami new APP         # Generate a new Hanami app
  hanami version         # Hanami version
```

### `hanami new`

```shell
⚡ hanami new --help
Command:
  hanami new

Usage:
  hanami new APP

Description:
  Generate a new Hanami app

Arguments:
  APP                               # REQUIRED App name

Options:
  --[no-]skip-install               # Skip app installation (Bundler, third-party Hanami plugins), default: false
  --help, -h                        # Print this help

Examples:
  hanami new bookshelf                # Generate a new Hanami app in `bookshelf/' directory, using `Bookshelf' namespace
  hanami new bookshelf --skip-install # Generate a new Hanami app, but it skips Hanami installation
```

### `hanami version`

```shell
⚡ hanami version --help
Command:
  hanami version

Usage:
  hanami version

Description:
  Hanami version

Options:
  --help, -h                        # Print this help
```

## App commands

Inside an application

### `hanami`

```shell
⚡ bundle exec hanami --help
Commands:
  hanami console                              # Start app console (REPL)
  hanami generate [SUBCOMMAND]
  hanami install                              # Install Hanami third-party plugins
  hanami middleware                           # Print app Rack middleware stack
  hanami routes                               # Print app routes
  hanami server                               # Start Hanami app server
  hanami version                              # Print Hanami app version
```

### `hanami console`

```shell
⚡ bundle exec hanami console --help
Command:
  hanami console

Usage:
  hanami console

Description:
  Start app console (REPL)

Options:
  --engine=VALUE                    # Console engine: (pry/irb)
  --help, -h                        # Print this help
```

### `hanami generate`

```shell
⚡ bundle exec hanami generate --help
Commands:
[...]

PR #92, opened 2023-08-25
Use Irb as default engine for hanami console.

nashby:
without this fix I'm getting following error when running `hanami console`:

```
From: /usr/local/bundle/gems/hanami-cli-2.0.3/lib/hanami/cli/commands/app/console.rb:49 Hanami::CLI::Commands::App::Command::Environment#call:

    44:               err.puts "`#{engine}' is not bundled. Please run `bundle add #{engine}' and retry."
    45:               exit(1)
    46:             end
    47:
    48:             console_engine.start
 => 49:           end
    50:
    51:           private
    52:
    53:           def resolve_engine(engine, opts)
    54:             if engine
```

### DIFF
--- a/lib/hanami/cli/commands/app/console.rb
+++ b/lib/hanami/cli/commands/app/console.rb
@@ -37,7 +37,7 @@ module Hanami
 
           # @since 2.0.0
           # @api private
-          def call(engine: nil, **opts)
+          def call(engine: DEFAULT_ENGINE, **opts)
             console_engine = resolve_engine(engine, opts)
 
             if console_engine.nil?
--- a/spec/unit/hanami/cli/commands/app/console_spec.rb
+++ b/spec/unit/hanami/cli/commands/app/console_spec.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+RSpec.describe Hanami::CLI::Commands::App::Console do
+  subject { described_class.new }
+
+  context "#call" do
+    context "with engine provided" do
+      let(:pry) do
+        instance_double(Hanami::CLI::Repl::Pry, name: "pry")
+      end
+
+      it "invokes pry" do
+        allow(Hanami::CLI::Repl::Pry).to receive(:new).and_return(pry)
+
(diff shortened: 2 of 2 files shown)


timriley:
Thanks for the report, @nashby! However, could you please provide some precise steps about how to replicate this?

I've tried myself by doing the following:

1. Generating a new Hanami app
2. Comment out `"hanami-reloader"` and `"guard-puma`" from the `Gemfile`
3. Run `bundle` again
4. Run `gem uninstall pry` to make sure that pry is not installed
5. Then inside the new app, run `bundle exec hanami console`

In this case, it successfully loads the console using IRB. Could you please let me know how I can get to this particular error you found?

Possibly sharing your installed gems would help here too.

My understanding is that irb is a [default gem](https://stdgems.org) and cannot be uninstalled, so in `Hanami::CLI::Commands::App::Console#resolve_engine`, we should always be able to rely on the final entry in the `ENTRIES` hash, which is irb.

timriley:
Depending on what we learn here, we may possible want to remove that `DEFAULT_ENGINE` constant, which is as of now unused (@jodosha, it looks like you added this in #53 — do you recall what your intention was here?)

nashby:
@timriley yep, sorry, false alarm. `pry-byebug` gem is the issue here. Not sure why but when hanami-cli starts `pry` this line is executed https://github.com/deivid-rodriguez/pry-byebug/blob/4eb7421103985215d404d42d3ed74cfec15f0102/lib/pry-byebug/pry_ext.rb#L12 Looks like they fixed this issue for Rails https://github.com/deivid-rodriguez/pry-byebug/pull/392 

I'll try to investigate more but this PR is not relevant for sure, closing.

timriley:
Wow, what an obscure combination of factors! Thanks so much for clarifying the cause, @nashby, and I'm glad to see it looks like the pry-byebug issue is being fixed too.

Please do feel free to loop back here if for any reason you think we might need to take mitigating measures inside Hanami 🙇🏼

PR #95, opened 2023-09-11
Make actions render their view files by default

### LINKED ISSUE #93: Update CLI action generators to allow rendering of paired view

radar:
Fixes #93

### DIFF
--- a/lib/hanami/cli/generators/app/action/action.erb
+++ b/lib/hanami/cli/generators/app/action/action.erb
@@ -4,8 +4,12 @@ module <%= camelized_app_name %>
   module Actions
 <%= module_controller_declaration %>
 <%= module_controller_offset %>class <%= camelized_action_name %> < <%= camelized_app_name %>::Action
-<%= module_controller_offset %>  def handle(*, response)
+<%- if hanami_view_bundled? -%>
+<%= module_controller_offset %>  def handle(_request, _response)
+<%- else -%>
+<%= module_controller_offset %>  def handle(_request, response)
 <%= module_controller_offset %>    response.body = self.class.name
+<%- end -%>
 <%= module_controller_offset %>  end
 <%= module_controller_offset %>end
 <%= module_controller_end %>
--- a/lib/hanami/cli/generators/app/action/action.erb
+++ b/lib/hanami/cli/generators/app/action/action.erb
@@ -7,7 +7,7 @@ module <%= camelized_app_name %>
 <%- if hanami_view_bundled? -%>
 <%= module_controller_offset %>  def handle(request, response)
 <%- else -%>
-<%= module_controller_offset %>  def handle(_request, response)
+<%= module_controller_offset %>  def handle(request, response)
 <%= module_controller_offset %>    response.body = self.class.name
 <%- end -%>
 <%= module_controller_offset %>  end
--- a/lib/hanami/cli/generators/app/action/action.erb
+++ b/lib/hanami/cli/generators/app/action/action.erb
@@ -5,7 +5,7 @@ module <%= camelized_app_name %>
 <%= module_controller_declaration %>
 <%= module_controller_offset %>class <%= camelized_action_name %> < <%= camelized_app_name %>::Action
 <%- if hanami_view_bundled? -%>
-<%= module_controller_offset %>  def handle(_request, _response)
+<%= module_controller_offset %>  def handle(request, response)
 <%- else -%>
 <%= module_controller_offset %>  def handle(_request, response)
 <%= module_controller_offset %>    response.body = self.class.name
--- a/lib/hanami/cli/generators/app/action/slice_action.erb
+++ b/lib/hanami/cli/generators/app/action/slice_action.erb
@@ -4,8 +4,12 @@ module <%= camelized_slice_name %>
   module Actions
 <%= module_controller_declaration %>
 <%= module_controller_offset %>class <%= camelized_action_name %> < <%= camelized_slice_name %>::Action
-<%= module_controller_offset %>  def handle(*, response)
+<%- if hanami_view_bundled?  -%>
+<%= module_controller_offset %>  def handle(_request, _response)
+<%- else -%>
+<%= module_controller_offset %>  def handle(_request, response)
 <%= module_controller_offset %>    response.body = self.class.name
+<%- end -%>
 <%= module_controller_offset %>  end
 <%= module_controller_offset %>end
 <%= module_controller_end %>
--- a/lib/hanami/cli/generators/app/action/slice_action.erb
+++ b/lib/hanami/cli/generators/app/action/slice_action.erb
@@ -7,7 +7,7 @@ module <%= camelized_slice_name %>
 <%- if hanami_view_bundled?  -%>
 <%= module_controller_offset %>  def handle(request, response)
 <%- else -%>
-<%= module_controller_offset %>  def handle(_request, response)
+<%= module_controller_offset %>  def handle(request, response)
 <%= module_controller_offset %>    response.body = self.class.name
 <%- end -%>
 <%= module_controller_offset %>  end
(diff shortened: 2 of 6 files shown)


jodosha (changes requested):
@radar Thanks for this PR, but there are a few things to change before merging it.

We should still generate a `#handle` signature because it serves as guidance for newcomers, and it's really like the case that after one generates an action, one wants to do something with it. Without the signature, a developer needs to write the method themselves. We want generators to be as much as possible useful.

My suggestion is to do:

```ruby
def handle(*)
end
```

---

The following line should appear or not, according to whether `hanami-view` is bundled or not.

```ruby
response.body = self.class.name
```

timriley:
@jodosha Given we want the handle method to be _informative_ to users, I feel we should probably generate both params for the method too:

```ruby
def handle(request, response)
```

That way at least users know what they're dealing with. They can always blank out unused params later if they want to.

I feel this will be pretty important to do for 2.1, otherwise _every_ action will get just `handle(*)` for a full stack app, which I think will eventually become annoying to users, especially given most actions will typically end up dealing with request params at minimum.

jodosha:
@timriley OK, maybe `req, res`? so it's shorter for people to type in the implementation of `#handle`.

timriley:
@jodosha we actually changed to the full name for this terms in the lead up to 2.0 itself, because it make the arguments less cryptic and more self explanatory. I think it'd be good if we could keep them expanded like this: typing a few letters more isn't all that hard, and it makes the resulting code much more expressive when _read_ (and it'll be read far more than it's written).

jodosha:
@radar there are specs failing, could you please have a look? Thanks 👍

radar:
@jodosha Please run the build again here. I've just pushed.

timriley (commented):
The changes here look good, thanks @radar!

I've left one comment about my preference for avoiding the leading underscore on the `request, response` parameter names. I'll see what @jodosha thanks, but after we have resolution on that, this looks good to go 🚀

lib/hanami/cli/generators/app/action/action.erb
```diff
@@ -4,8 +4,12 @@ module <%= camelized_app_name %>
   module Actions
 <%= module_controller_declaration %>
 <%= module_controller_offset %>class <%= camelized_action_name %> < <%= camelized_app_name %>::Action
-<%= module_controller_offset %>  def handle(*, response)
+<%- if hanami_view_bundled? -%>
+<%= module_controller_offset %>  def handle(_request, _response)
```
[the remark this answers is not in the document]
jodosha: @timriley Yes.

lib/hanami/cli/generators/app/action/action.erb
```diff
@@ -4,8 +4,12 @@ module <%= camelized_app_name %>
   module Actions
 <%= module_controller_declaration %>
 <%= module_controller_offset %>class <%= camelized_action_name %> < <%= camelized_app_name %>::Action
-<%= module_controller_offset %>  def handle(*, response)
+<%- if hanami_view_bundled? -%>
+<%= module_controller_offset %>  def handle(request, response)
+<%- else -%>
+<%= module_controller_offset %>  def handle(_request, response)
```
jodosha, line 10:
```suggestion
<%= module_controller_offset %>  def handle(request, response)
```

lib/hanami/cli/generators/app/action/slice_action.erb
```diff
@@ -4,8 +4,12 @@ module <%= camelized_slice_name %>
   module Actions
 <%= module_controller_declaration %>
 <%= module_controller_offset %>class <%= camelized_action_name %> < <%= camelized_slice_name %>::Action
-<%= module_controller_offset %>  def handle(*, response)
+<%- if hanami_view_bundled?  -%>
+<%= module_controller_offset %>  def handle(_request, _response)
```
jodosha, line 8:
```suggestion
<%= module_controller_offset %>  def handle(request, response)
```

lib/hanami/cli/generators/app/action/slice_action.erb
```diff
@@ -4,8 +4,12 @@ module <%= camelized_slice_name %>
...
+<%- else -%>
+<%= module_controller_offset %>  def handle(_request, response)
```
jodosha, line 10:
```suggestion
<%= module_controller_offset %>  def handle(request, response)
```

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -51,7 +51,7 @@ module #{inflector.camelize(app)}
             module Actions
               module #{inflector.camelize(controller)}
                 class #{inflector.camelize(action)} < #{inflector.camelize(app)}::Action
-                  def handle(*, response)
+                  def handle(_request, response)
```
jodosha, line 54:
```suggestion
                  def handle(request, response)
```

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -182,7 +182,7 @@ module Actions
               module API
                 module Users
                   class Thing < #{inflector.camelize(app)}::Action
-                    def handle(*, response)
+                    def handle(_request, response)
```
jodosha, line 185:
```suggestion
                    def handle(request, response)
```

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -226,6 +226,114 @@ def handle(*, response)
...
+
+        # action
+        action_file = <<~EXPECTED
+          # frozen_string_literal: true
+
+          module #{inflector.camelize(app)}
+            module Actions
+              module #{inflector.camelize(controller)}
+                class #{inflector.camelize(action)} < #{inflector.camelize(app)}::Action
+                  def handle(_request, _response)
```
jodosha, line 263:
```suggestion
                  def handle(request, response)
```

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -226,6 +226,114 @@ def handle(*, response)
...
+
+        action_file = <<~EXPECTED
+          # frozen_string_literal: true
+
+          module #{inflector.camelize(app)}
+            module Actions
+              module API
+                module Users
+                  class Thing < #{inflector.camelize(app)}::Action
+                    def handle(_request, _response)
```
jodosha, line 322:
```suggestion
                    def handle(request, response)
```

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -270,7 +378,7 @@ module #{inflector.camelize(slice)}
             module Actions
               module #{inflector.camelize(controller)}
                 class #{inflector.camelize(action)} < #{inflector.camelize(slice)}::Action
-                  def handle(*, response)
+                  def handle(_request, response)
```
jodosha, line 381:
```suggestion
                  def handle(request, response)
```

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -360,7 +468,7 @@ module Books
                   module Bestsellers
                     module Nonfiction
                       class #{inflector.camelize(action)} < #{inflector.camelize(slice)}::Action
-                        def handle(*, response)
+                        def handle(_request, response)
```
jodosha, line 471:
```suggestion
                        def handle(request, response)
```

jodosha (approved):

jodosha:
I couldn't commit all the suggestions in a batch for some reason.

Screenshot 2023-09-18 at 09 11 40

jodosha:
@radar Thank you very much. 👍

PR #98, opened 2023-09-11
Use app's configured assets exe_path for commands

### LINKED ISSUE #96: Hanami Assets CLI: get the path of the NPM executable from config
See:
* https://github.com/hanami/cli/blob/086a69e50e03d7c92c2f069ac55374591ec64a31/lib/hanami/cli/commands/app/assets/compile.rb#L18
* https://github.com/hanami/cli/blob/086a69e50e03d7c92c2f069ac55374591ec64a31/lib/hanami/cli/commands/app/assets/watch.rb#L23

timriley:
For the `assets compile` and `assets watch` commands, instead of hardcoding a default path to the `hanami-assets.js` command, use the configured exe path at `Hanami.app.assets.exe_path`.

Resolves #96.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,9 +10,10 @@ unless ENV["CI"]
 end
 
 gem "hanami", github: "hanami/hanami", branch: "main"
-gem "hanami-utils", github: "hanami/utils", branch: "main"
-gem "hanami-router", github: "hanami/router", branch: "main"
+gem "hanami-assets", github: "hanami/assets", branch: "main"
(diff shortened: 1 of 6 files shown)


jodosha (approved):
@timriley This is going on the right direction: take the `exe_path` from `hanami-assets`.

I'll open a new PR to rely on `npm exec`, instead of configuring the exact path of the executable. The path is an NPM implementation detail that may change in the future. Better to rely on their Public API command line.

PR #99, opened 2023-09-11
App and slice generators for assets

jodosha:
# Enhancements

Enhance `hanami new` and `hanami generate slice` commands to support assets.

# Technical Implementation

## Skip Assets

By default, assets are added to a newly created 2.1 app.

We introduced a new option `--[no-]skip-assets` to be used like this:

```bash
$ hanami new bookshelf --skip-assets
```

With this option, a developer can opt-out of `hanami-assets`.

NOTE: This option opens the possibility to do the same for other Hanami gems.

## `hanami install`

We expanded the behavior of this command to setup NPM and install `hanami-assets` NPM package.

## `Gemfile`

By default, it adds the `hanami-assets` gem to the app's `Gemfile`.

## `Procfile.dev`

It generates a `Procfile.dev` file, which is used by the new command `hanami dev` (see https://github.com/hanami/cli/pull/97).

## Assets

It generates:

  * `app/assets/javascripts/app.js`
  * `app/assets/stylesheets/app.css`
  * `app/assets/images/favicon.ico`

## App layout

It generates the app layout with basic HTML5 markup and a `stylesheet` tag.

## Gitignore

If `hanami-assets` is present, it adds the following entries to the default `.gitignore`:

```
public/
node_modules/
```

---

## Slices

Generate assets artifacts, only if `hanami-assets` is bundled.

### App layout

It generates the app layout with basic HTML5 markup and a `stylesheet` tag.

### Assets

It generates:

  * `slices/admin/assets/javascripts/app.js`
  * `slices/admin/assets/stylesheets/app.css`


lib/hanami/cli/generators/gem/app/gitignore.erb
```diff
@@ -1,2 +1,6 @@
 .env
 log/*
+<%- if Hanami.bundled?("hanami-assets") -%>
+public/
```
timriley, line 4: Question: should we be checking in a public/.keep to make sure this directory exists in the first place?

lib/hanami/cli/generators/gem/app/app_layout.erb
```diff
@@ -0,0 +1,15 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title><%= humanized_app_name %></title>
+    <%- if bundled_assets? -%>
+    <%%= favicon %>
+    <%- end -%>
+  </head>
+  <body>
+    <%%= yield %>
```
timriley, line 14: Nice pickup with needing `<%%=` here. Thank you! 🙇🏼

lib/hanami/cli/generators/gem/app/app_layout.erb
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
```
timriley, line 6: Could we drop this line? Seems quite outmoded now. [See here](https://stackoverflow.com/a/6771584).

lib/hanami/cli/generators/gem/app/app_js.erb
```diff
@@ -0,0 +1 @@
+import "../stylesheets/app.css";
```
timriley, line 1: Forgive me if this is an ignorant question... but why do we need to import the app's _stylesheet_ from inside this JavaScript file?

lib/hanami/cli/generators/app/slice_context.rb
```diff
@@ -29,6 +29,18 @@ def underscored_slice_name
...
+          # @since 2.1.0
+          # @api private
+          def humanized_slice_name
+            inflector.humanize(slice)
+          end
+
+          # @since 2.1.0
+          # @api private
+          def stylesheet_erb_tag
+            %(<%= stylesheet "#{slice}/app" %>)
```
timriley, line 41: Penny for your thoughts: what do you think about seeing a file named "app" inside a _slice_?

(This is absolutely not intended to block this PR from merging, but I figure now is a good time to briefly consider/confirm our naming choices).

lib/hanami/cli/generators/gem/app.rb
```diff
@@ -54,7 +55,13 @@ def generate_app(app, context) # rubocop:disable Metrics/AbcSize
             fs.write("app/action.rb", t("action.erb", context))
             fs.write("app/view.rb", t("view.erb", context))
             fs.write("app/views/helpers.rb", t("helpers.erb", context))
-            fs.write("app/templates/layouts/app.html.erb", File.read(File.join(__dir__, "app", "layouts_app.html.erb")))
+            fs.write("app/templates/layouts/app.html.erb", t("app_layout.erb", context))
+
+            if context.generate_assets?
+              fs.write("app/assets/javascripts/app.js", t("app_js.erb", context))
+              fs.write("app/assets/stylesheets/app.css", t("app_css.erb", context))
```
timriley, line 62: What do you think about using `/js/` and `/css/` dirs instead of these rather long names? That would fit well with our short `js` and `css` asset helpers.

(This is not critical, but I figure now is as good a time as any to consider these names)

lib/hanami/cli/generators/app/slice/app_layout.erb
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title><%= humanized_app_name %> - <%= humanized_slice_name %></title>
+    <%- if bundled_assets? -%>
+    <%%= favicon %>
```
timriley, line 9: We'll need to add this back as an alias to Hanami::Helpers::AssetsHelper, which right now only has `#favicon_link_tag`. I can do that in a PR after this.

spec/unit/hanami/cli/commands/gem/new_spec.rb
```diff
@@ -1,15 +1,22 @@
 # frozen_string_literal: true
 
 RSpec.describe Hanami::CLI::Commands::Gem::New do
-  subject {
+  subject do
     described_class.new(bundler: bundler, out: out, fs: fs, inflector: inflector)
-  }
+  end
+
+  # TODO: remove this block
```
timriley, line 8: Just checking that you intend for this TODO to be kept here?

lib/hanami/cli/generators/app/slice/app_layout.erb
```diff
@@ -0,0 +1,16 @@
...
+    <%= stylesheet_erb_tag %>
```
timriley, line 10: Do we also want to generate a default tag to include the main app JS file?

timriley (approved):
Looks great, @jodosha!

I have left a bunch of different notes/questions for you, but there's nothing there that I think is worth blocking a merge. Definitely keen for your thoughts on them, though :)

lib/hanami/cli/generators/gem/app/app_layout.erb
jodosha: Note for the future: because this is an ERB template that needs to output a `.erb` file if we don't use the escape `<%%=`, we can't output `<%= yield %>`. The ERB template of the code generators will attempt to yield immediately.

lib/hanami/cli/generators/gem/app/app_layout.erb
jodosha: @timriley Cargo-culting at its finest. Sorry about that.
Removed: https://github.com/hanami/cli/pull/99/commits/abbb927ae1dbc2aaede98d78adf43910d86aebd3

lib/hanami/cli/generators/gem/app/app_js.erb
jodosha: @timriley This is because Esbuild is only processing assets (JS and CSS) referenced from our entry points.

Entry points are:
* `app/assets/javascripts/**/app.js`
* `slices/*/assets/javascripts/**/app.js`

We support many other extensions: TS, MJS, JSX, etc...

Esbuild does that for efficiency reasons: only process (and bundle) the required code.

---

We have two options:

1. (Current) Take Esbuild defaults. This means that CSS needs to be referenced directly (or indirectly) from entry points. That's why we import the CSS from that file.
2. Implement a custom strategy for CSS, just always process ALL the stylesheets ALL the time. Like we do for static assets.

lib/hanami/cli/generators/app/slice_context.rb
jodosha: @timriley I followed the convention of naming "app" those main artifacts. Similar to `slice/admin/layouts/app.html.erb`.

lib/hanami/cli/generators/gem/app.rb
jodosha: @timriley Good one that could work.

To your knowledge, this requires a change in `hanami/hanami-js` (the Esbuild plugin), which is doable.
Not complaining about work, just sharing awareness. 😜

lib/hanami/cli/generators/app/slice/app_layout.erb
jodosha: @timriley Whoa, you were faster than my request. 😂 Thanks.

spec/unit/hanami/cli/commands/gem/new_spec.rb
jodosha: @timriley Good catch: removed. https://github.com/hanami/cli/pull/99/commits/c89e6f84263ba6c076af9cf198e0113752d2acae

lib/hanami/cli/generators/gem/app.rb
jodosha: @timriley Change in this PR: https://github.com/hanami/cli/pull/99/commits/264782eac857460a0e6d4b4b0c96353fae60ac5c

jodosha: @timriley Done.

lib/hanami/cli/generators/app/slice/app_layout.erb
jodosha: @timriley Yes, please see https://github.com/hanami/cli/pull/99/commits/ede966a514ba1b181c95ad7acccfa882dca5599e

jodosha:
@timriley, What's your opinion about the asset helpers generated with the application template?

1. `<%= javascript "app" %>` and `<%= stylesheet "app" %>` (current)
2.`<%= js "app" %>` and `<%= css "app" %>`

---

I vote for something consistent with the naming of the directories.

1. `app/assets/javascripts` should go with `<%= javascript %>`
2.`app/assets/js` should go with `<%= js %>`

What do you think?

lib/hanami/cli/generators/gem/app/app_js.erb
timriley: Thanks for clarifying this for me, @jodosha!

I think that following Esbuild's defaults is probably the most sensible option for us now. Let's roll with this and ship it with the next beta, then decide after that whether we might need to directly process CSS files?

timriley: To help with our eventual decision here, it might be good to come up with a range of different "personas" for asset structures (e.g. a simple site with CSS only but no JS, a site with React and Tailwind, and maybe some third structure) — then we can survey how well we serve each scenario with our current approach vs an approach with direct CSS handling. What do you reckon?

lib/hanami/cli/generators/app/slice_context.rb
timriley: This makes sense! Let's stick with it 👍🏼

timriley:
@jodosha regarding the helper names and directories to use, I will share one thing, which may be irrational, but so be it! I never liked the term "javascripts" — IMO, this is supremely awkward pluralisation! No one would ever use that term in natural speech.

For this reason alone, I would vote `js/` and `css/` as directories, and the matching short helper names in the default layout `(js` and `css`).

However, in the cases where we also have `images/` and `fonts/` directories, I do realise that this would make these names inconsistent: "images" e.g. being a full term, pluralised, whereas "js" and "css" are singular abbreviations.

However, I think js and css are special due to their important foundational roles in delivering front end features, so I think it's OK for their directories to be somewhat special too.

Also, elixir uses assets/js/, so we're not going off on our own here.

OK, so that's a lot of words, but now you know the entirety of my thinking here :)

PR #108, opened 2023-10-06
`hanami new` to generate documented `config/puma.rb`

jodosha:
Closes https://github.com/hanami/hanami/issues/1340

### DIFF
--- a/lib/hanami/cli/generators/gem/app/puma.erb
+++ b/lib/hanami/cli/generators/gem/app/puma.erb
@@ -35,12 +35,13 @@ if puma_cluster_mode
   # Preload the application before starting the workers. Only in cluster mode.
   preload_app!
 
-  # Code to run immediately before a worker shuts down, after it has finished
-  # processing HTTP requests.
+  # Code to run immediately before master process forks workers (once on boot).
   #
   # These hooks can block if necessary to wait for background operations unknown
-  # to Puma to finish before the process terminates.
-  on_worker_shutdown do
+  # to puma to finish before the process terminates. This can be used to close
+  # any connections to remote servers (database, redis, …) that were opened when
+  # preloading the code.
+  before_fork do
     Hanami.shutdown
   end
 end
--- a/lib/hanami/cli/generators/gem/app/puma.erb
+++ b/lib/hanami/cli/generators/gem/app/puma.erb
@@ -20,18 +20,18 @@ threads min_threads_count, max_threads_count
 # Workers (aka Puma/Ruby processes)
 #
 
-hanami_concurrency = ENV.fetch("HANAMI_WEB_CONCURRENCY", 0)
-hanami_cluster_mode = hanami_concurrency.positive?
+puma_concurrency = ENV.fetch("HANAMI_WEB_CONCURRENCY", 0)
+puma_cluster_mode = puma_concurrency.positive?
 
 # How many worker (Puma/Ruby) processes to run.
 # Typically this is set to the number of available cores.
-workers hanami_concurrency
+workers puma_concurrency
 
 #
 # Cluster mode (aka multiple workers)
 #
 
-if hanami_cluster_mode
+if puma_cluster_mode
   # Preload the application before starting the workers. Only in cluster mode.
   preload_app!
 
--- a/lib/hanami/cli/generators/gem/app/puma.erb
+++ b/lib/hanami/cli/generators/gem/app/puma.erb
@@ -1,17 +1,46 @@
 # frozen_string_literal: true
 
-max_threads_count = ENV.fetch("HANAMI_MAX_THREADS", 5)
-min_threads_count = ENV.fetch("HANAMI_MIN_THREADS") { max_threads_count }
-threads min_threads_count, max_threads_count
-
+#
+# Environment and port
+#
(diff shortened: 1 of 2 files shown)


timriley:
@jodosha I've just pushed up a range of small fixes/tweaks (all the initial commits), as well as a suggested adjustment for the hook we use for running `Hanami.shutdown` (the last commit).

I've left explanations for each in the extended commit messages. Let me know what you think 😄 

In terms of verification of the Puma `before_fork` usage, I've tested this with my local copy of decafsucks using this `config/puma.rb`, and the following `persistence` provider (note the `stop` lifecycle method):

```ruby
# frozen_string_literal: true

Hanami.app.register_provider :persistence, namespace: true do
  prepare do
    require "rom-changeset"
    require "rom/core"
    require "rom/sql"

    # TODO(Hanami): As part of built-in rom setup, configure ROM with app inflector
    ROM::Inflector = Hanami.app["inflector"]

    rom_config = ROM::Configuration.new(:sql, target["settings"].database_url)

    rom_config.plugin(:sql, relations: :instrumentation) do |plugin_config|
      plugin_config.notifications = target["notifications"]
    end

    rom_config.plugin(:sql, relations: :auto_restrictions)

    register "config", rom_config
    register "db", rom_config.gateways[:default].connection
  end

  start do
    rom_config = target["persistence.config"]
    rom_config.auto_registration(
      target.root.join("app"),
      namespace: Hanami.app.namespace.to_s
    )

    register "rom", ROM.container(rom_config)
  end

  stop do
    target["persistence.rom"].disconnect
  end
end
```

Then I run the Hanami server like so: `HANAMI_WEB_CONCURRENCY=2 be hanami server`

This shows Puma entering cluster mode as expected:

```
22:28:55 - INFO - Using Guardfile at /Users/tim/Source/decafsucks/decafsucks/Guardfile.
22:28:55 - INFO - Puma starting on port 2300 in development environment.
22:28:55 - INFO - Guard is now watching at '/Users/tim/Source/decafsucks/decafsucks'
[15180] Puma starting in cluster mode...
[15180] * Puma version: 6.4.0 (ruby 3.2.2-p53) ("The Eagle of Durango")
[15180] *  Min threads: 5
[15180] *  Max threads: 5
[15180] *  Environment: development
[15180] *   Master PID: 15180
[15180] *      Workers: 2
[15180] *     Restarts: (✔) hot (✖) phased
[15180] * Preloading application
[15180] * Listening on http://0.0.0.0:2300
[15180] Use Ctrl-C to stop
[15180] * Starting control server on http://127.0.0.1:9293
[15180] * Starting control server on http://[::1]:9293
[15180] - Worker 0 (PID: 15183) booted in 0.0s, phase: 0
[15180] - Worker 1 (PID: 15184) booted in 0.0s, phase: 0
```

And then I make a request to a page that hits the database, and everything still works as expected.

So I think we're good to go on this front for now. 👍🏼 When we eventually do get to integrating rom for real in 2.2, we might want to adjust our `on_worker_boot` code to be more specific, but for now I think this is a good arrangement.

timriley:
@jodosha Ah! I did not know about puma casting integers internally. So that means we only need to do it when we want to use a variable as an integer elsewhere in this file. Thanks for the pointer 👍🏼 

As for the `puma_cluster_mode`, I actually tried it with 1, via `HANAMI_WEB_CONCURRENCY=1 bundle exec hanami server`, and got this big fat warning in the output (blank lines added for emphasis):

```
23:32:50 - INFO - Using Guardfile at /Users/tim/Source/decafsucks/decafsucks/Guardfile.
23:32:50 - INFO - Puma starting on port 2300 in development environment.
23:32:50 - INFO - Guard is now watching at '/Users/tim/Source/decafsucks/decafsucks'
[16483] Puma starting in cluster mode...
[16483] * Puma version: 6.4.0 (ruby 3.2.2-p53) ("The Eagle of Durango")
[16483] *  Min threads: 5
[16483] *  Max threads: 5
[16483] *  Environment: development
[16483] *   Master PID: 16483
[16483] *      Workers: 1
[16483] *     Restarts: (✔) hot (✔) phased
[16483] * Listening on http://0.0.0.0:2300
[16483] Use Ctrl-C to stop

[16483] ! WARNING: Detected running cluster mode with 1 worker.
[16483] ! Running Puma in cluster mode with a single worker is often a misconfiguration.
[16483] ! Consider running Puma in single-mode (workers = 0) in order to reduce memory overhead.
[16483] ! Set the `silence_single_worker_warning` option to silence this warning message.

[16483] * Starting control server on http://127.0.0.1:9293
[16483] * Starting control server on http://[::1]:9293
[16483] - Worker 0 (PID: 16485) booted in 0.35s, phase: 0
```

It was for the sake of avoiding this warning that I changed it to `> 1` instead of `.positive?`.

I feel in two minds about this:

1. I think it's good for to use our defaults to steer people away from potentially nonsensical configurations (like cluster mode with a worker count of 1, which is exactly why Puma prints that kind of warning)
2. On the other hand, I also don't want our defaults to make certain configurations _impossible_, especially if the user is being intentional about it.

Given both of these, I felt like `> 1` strikes a reasonable balance, because a user who _really_ wants to run cluster mode with 1 worker can simply change that line in their own config to `> 0` or `.positive?` and then get on with their life.

What do you think?

jodosha:
@timriley I double checked: yes, you're right, with `workers` we I get a warning too.

PR #109, opened 2023-10-11
Generate and install new assets arrangement

timriley:
Prepare for our updated approach to assets chiefly introduced with https://github.com/hanami/assets-js/pull/10.

Update `hanami new`:

- Generate a new `config/assets.mjs` that will be needed for our new way of running the assets commands (see https://github.com/hanami/assets-js/pull/10)
- - Generate a leaner `package.json`, including an `"assets"` entry in the `"scripts"` section, referencing the above `config/assets.mjs` file.

Update the `hanami assets compile` and `hanami assets watch` commands:

- Expect to run assets via `npm run assets`, as generated in the new `package.json` above.
- Have these use a single hanami-assets setting for running these commands: `package_manager_run_command` (introduced in https://github.com/hanami/assets/pull/131), which defaults to `npm run`.

### DIFF
--- a/lib/hanami/cli/generators/gem/app/assets.mjs
+++ b/lib/hanami/cli/generators/gem/app/assets.mjs
@@ -1,0 +1,14 @@
+import * as assets from "hanami-assets";
+
+await assets.run();
+
+// To provide additional esbuild (https://esbuild.github.io) options, use the following:
+//
+// await assets.run({
+//   esbuildOptionsFn: (args, esbuildOptions) => {
+//     // Add to esbuildOptions here. Use `args.watch` as a condition for different options for
+//     // compile vs watch.
+//
+//     return esbuildOptions;
+//   }
+// });
--- a/lib/hanami/cli/generators/gem/app/package.json.erb
+++ b/lib/hanami/cli/generators/gem/app/package.json.erb
@@ -4,7 +4,7 @@
   "scripts": {
     "assets": "node config/assets.mjs"
   },
-  "devDependencies": {
+  "dependencies": {
     <%= hanami_assets_npm_package %>
   }
 }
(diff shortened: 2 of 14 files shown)


lib/hanami/cli/generators/gem/app/package.json.erb
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "<%= underscored_app_name %>",
+  "private": true,
+  "scripts": {
+    "assets": "node config/assets.mjs"
+  },
+  "devDependencies": {
```
jodosha, line 7:
```suggestion
  "dependencies": {
```

@timriley We need the NPM package to be installed in production as well for deployment reasons.

lib/hanami/cli/generators/gem/app/assets.mjs
```diff
@@ -0,0 +1,14 @@
+import * as assets from "hanami-assets";
+
+await assets.run();
+
+// To provide additional esbuild (https://esbuild.github.io) options, use the following:
```
jodosha, line 5: @timriley Good for now. We'll need to replace this code snipped with a link to a future Hanami Guide article. 😉

jodosha (changes requested):
@timriley This looks great. 👍 

The only reason why I'm rejecting this is that we'll probably need to change the way we generate `package.json`. Please see my inline comment.

lib/hanami/cli/generators/gem/app/package.json.erb
timriley: Ohhh, great pickup, thank you! Fixed now.

lib/hanami/cli/generators/gem/app/assets.mjs
timriley: Good call. Have made [a draft issue](https://github.com/orgs/hanami/projects/2/views/1?pane=issue&itemId=41340063) in our project to remind us.

jodosha (approved):

PR #123, opened 2023-11-05
Remove unused assets settings

timriley:
We're not using the `entry_points` and `destination` assets settings right now. These are currently hardcoded inside assets-js, so remove these references to avoid any confusion that we may be able to configure these from the outside. See https://github.com/hanami/assets/pull/132, where we're removing those settings directly from the hanami-assets gem.

While we're here, properly shellescape the `sri` argument given to the assets CLI.


jodosha (approved):
@timriley 👍

PR #126, opened 2023-11-06
Add --skip-tests flag to generate commands

timriley:
This will allow hanami-rspec to observe this flag and skip its work if true.

This is in service of making sure we don't generate spurious (and failing) tests in our new 2.1 getting started guide for web apps.

(More generally, it would be nice if gems adding hooks around the core CLI commands could register their own options, since that would mean we could limit these changes to the hanami-rspec gem only, but for now, we have to make this change to register the flag here, and then use the flag over there)

### DIFF
--- a/lib/hanami/cli/commands/app/generate/action.rb
+++ b/lib/hanami/cli/commands/app/generate/action.rb
@@ -22,12 +22,25 @@ module Hanami
             DEFAULT_SKIP_VIEW = false
             private_constant :DEFAULT_SKIP_VIEW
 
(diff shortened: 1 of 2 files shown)


jodosha:
@timriley 
> (More generally, it would be nice if gems adding hooks around the core CLI commands could register their own options, since that would mean we could limit these changes to the hanami-rspec gem only, but for now, we have to make this change to register the flag here, and then use the flag over there)

Good point. This is an enhancement for `dry-cli`.

jodosha (approved):

PR #128, opened 2023-11-07
Run bundle check || bundle install after hanami install

cllns:
To ensure any new Gemfile entries are installed


timriley (approved):
I'm really glad you spotted this, @cllns! Fix looks good to me.

@jodosha — I think this would be good to get in before 2.1.0 stable, but I don't mind when. If this has to wait until after rc2, I'm comfortable with that.

Let us know if/when you think this might be good to merge.

jodosha (approved):
@cllns Thanks for spotting and fixing this. It adds a new level of solidity to the installation. 🙏

PR #130, opened 2023-11-24
Implement generic component generator

krzykamil:
### **Context:**
Hi. I noticed this task in Hanami github projects: https://github.com/orgs/hanami/projects/2?pane=issue&itemId=37953496
Thought I give it a go, to get acquainted with dry-rb and hanami source code more, and maybe be useful to maintainers as a side effect. I know the task was marked as "maybe" but I often wrote generators for rails and wanted to see how much different would it be here.

### **Solution:**
The generator is really simple, as components are general objects with no one interface or anything. It is basically a shortcut to creating a file with the right name in the right scope and slice or no slice. That is what I understood that the idea was behind this task.
I also accounted for CLI receiving the name in Pascal Case like `Opeations::CreateUser`. This is a personal preference, since this is how I usually generate new ruby classes through rubymine, but I think it might be a useful option for some. 

### **TBD:**
For my code, the collection of constants with different characters used in `lib/hanami/cli/generators/app/view_context.rb` was also useful. I noticed the TODO that it should be moved. I would like to do thath, but I am unsure what would be the desired location. Something like `lib/hanami/cli/constants.rb` seems like a simple solution? But I was thinking more along the lines of `lib/hanami/cli/formatting` since all those constants are used to format the input or output, however that does not seem 100% right (this naming seems like a stretch, or might be misleading, since it is not functional, it would just be a module with constants or something).

Anyway, eager to see any form of feedback. Would you like the generator to do something more? Any feedback is appreciated since this is my first attempt at hanami source code.

I was also unsure about how to document the code with the version tags, since I started this before 2.1 was released.

Some screenshots of usage of my local gem version:

cli_component_slice
cli_component

### DIFF
--- a/lib/hanami/cli/commands/app/generate/component.rb
+++ b/lib/hanami/cli/commands/app/generate/component.rb
@@ -9,6 +9,7 @@ module Hanami
       module App
         module Generate
           # @api private
+          # @since 2.2.0
           class Component < App::Command
             argument :name, required: true, desc: "Component name"
             option :slice, required: false, desc: "Slice name"
@@ -16,13 +17,14 @@ module Hanami
             example [
               %(operations.create_user               (MyApp::Operations::CreateUser)),
               %(operations.user.create               (MyApp::Operations::Create::User)),
-              %(operations.create_user --slice=admin (Admin::Operations::CreateUser)),
-              %(Operations::CreateUser (MyApp::Operations::CreateUser)),
+              %(operations.create_user               --slice=admin (Admin::Operations::CreateUser)),
+              %(Operations::CreateUser               (MyApp::Operations::CreateUser)),
             ]
             attr_reader :generator
             private :generator
 
             # @api private
+            # @since 2.2.0
             def initialize(
               fs: Hanami::CLI::Files.new,
               inflector: Dry::Inflector.new,
@@ -34,6 +36,7 @@ module Hanami
             end
 
             # @api private
+            # @since 2.2.0
             def call(name:, slice: nil, **)
               slice = inflector.underscore(Shellwords.shellescape(slice)) if slice
 
--- a/lib/hanami/cli/commands/app/generate/component.rb
+++ b/lib/hanami/cli/commands/app/generate/component.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+require "dry/inflector"
+require "dry/files"
+require "shellwords"
+module Hanami
+  module CLI
+    module Commands
+      module App
+        module Generate
+          # @api private
+          class Component < App::Command
+            argument :name, required: true, desc: "Component name"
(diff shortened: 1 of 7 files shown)


krzykamil:
https://github.com/hanami/cli/issues/145

lib/hanami/cli/commands/app/generate/component.rb
```diff
@@ -0,0 +1,47 @@
...
+          # @api private
+          class Component < App::Command
+            argument :name, required: true, desc: "Component name"
+            option :slice, required: false, desc: "Slice name"
+
+            example [
+              %(operations.create_user               (MyApp::Operations::CreateUser)),
+              %(operations.user.create               (MyApp::Operations::Create::User)),
+              %(operations.create_user --slice=admin (Admin::Operations::CreateUser)),
+              %(Operations::CreateUser (MyApp::Operations::CreateUser)),
```
timriley, line 20: Should this line be spaced out so `(MyApp::Operations::CreateUser)` aligns with the lines above?

lib/hanami/cli/generators/app/component_context.rb
```diff
@@ -0,0 +1,84 @@
...
+          private_constant :MATCHER_PATTERN
+
+          NAMESPACE_SEPARATOR = "::"
+          private_constant :NAMESPACE_SEPARATOR
+
+          INDENTATION = "  "
+          private_constant :INDENTATION
+
+          OFFSET = 1
+          private_constant :OFFSET
```
timriley, line 19: For starters, let's try moving these to a generators/constants.rb, and make sure we have a manual require that ensures this file is loaded where necessary?

I'd be very happy to have this PR merge first, however, and then have you tackle that as a follow-up :)

timriley (commented):
This all looks great to me, @krzykamil — it's exactly what I had in mind!

Tidy up suggestions: we can remove the `# @private` comment from methods that already have private visibility (I know we're a bit inconsistent with this, but I'd like us to lean more in the direction of not having spurious API docs).

As for the public things (classes and public methods), let's add:

```
@api private
@since 2.2.0
```

Once you've done that, I think this is good to merge!

lib/hanami/cli/generators/app/component_context.rb
krzykamil: Yes, I think it should be done separately

lib/hanami/cli/commands/app/generate/component.rb
krzykamil: corrected the indentation

krzykamil:
@timriley I removed `private` from methods that are already in `private` scope and changed the version to 2.2.0 on stuff that was added.

timriley:
Thanks @krzykamil! 😍

PR #160, opened 2024-06-02
Update `db migrate` for new db layer

### LINKED ISSUE #151: Add `db migrate` command
```
$ bundle exec hanami db migrate # Migrates all databases
$ bundle exec hanami db migrate --app # Migrates app database
$ bundle exec hanami db migrate --slice=billing # Migrates "billing" slice database
```

Note: this already exists as a command, but it has been disabled for a while. It just needs to be reactivated, and updated for the per-slice behaviour.

timriley:
Update `db migrate` to work for multiple databases configured across the app and its slices.

To find databases to operate on, it:

1. Looks for all slices (app included) with a `:db` provider present, and groups these by their database URLs
2. For each database URL, finds the slice that has a `config/db/` directory - this is considered the "canonical" slice for the database. 
3. Migrates the database for each such slice.
    - If >1 slices have a config/db/, it prints a warning, and choose to operate on the first slice only.

The command also supports the following arguments:

- `--app` to operate on the app's db only
- `--slice=slice_name` to operate on a specific slice's database (in this case, we operate directly on that slice, and don't take the steps noted above to try and find the canonical slice for the database; we trust that the user has provided the right slice via this argument)

Resolves #151

### DIFF
--- a/lib/hanami/cli/commands/app/db/command.rb
+++ b/lib/hanami/cli/commands/app/db/command.rb
@@ -78,6 +78,7 @@ module Hanami
               elsif slices.length < 1
                 out.puts <<~STR
                   WARNING: Database #{database.name} has no config/db/ directory.
+
                 STR
               end
             end
--- a/lib/hanami/cli/commands/app/db/command.rb
+++ b/lib/hanami/cli/commands/app/db/command.rb
@@ -48,23 +48,38 @@ module Hanami
                 slice.prepare :db
                 database_url = slice["db.gateway"].connection.uri
 
-                hsh[database_url] ||= [] << slice
+                hsh[database_url] ||= []
+                hsh[database_url] << slice
               }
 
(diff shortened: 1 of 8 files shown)


lib/hanami/cli/commands/app/db/command.rb
```diff
@@ -0,0 +1,89 @@
...
+
+                database = Utils::Database[slices_with_config.first || slices.first]
+
+                warn_on_misconfigured_database database, slices_with_config
+
+                arr << database
+              }
+
+              databases
+            end
```
timriley, line 66: This is a pretty important method. We'll likely want something like this for all our `db` subcommands, which is why I've put it here.

As we begin to use this from the other subcommands, I expect we may want to tweak it a little. We'll see.

This is also why I didn't try and factor this out into smaller methods. Each time I tried to do so, it felt like it impaired the ease of understanding exactly what's going on to prepare this list of all databases, so I've left it as a single large-ish method for now.

PR #162, opened 2024-06-07
Adapt component generator to new command defaults

krzykamil:
This adapts to https://github.com/hanami/cli/pull/137/files which was created after I created https://github.com/hanami/cli/pull/130 but merged before it. My branch wasn't rebased before merge and since it was new files, there was no conflict. 

This change simply adapts to what the parent object does now withi default initialize params.


timriley:
Thanks @krzykamil, nice work figuring this out!

PR #163, opened 2024-06-11
Enable `db structure dump` for new db layer

### LINKED ISSUE #154: Add `db structure dump` command
```
$ bundle exec hanami db structure dump # Dumps the structure for all databases
$ bundle exec hanami db structure dump --app # Dumps the structure for the app database
$ bundle exec hanami db structure dump --slice=billing # Dumps the structure for the "billing" slice database
```

Note: this already exists as a command, but it has been disabled for a while. It just needs to be reactivated, and updated for the per-slice behaviour.

timriley:
Enable the `db structure dump` command for our new database layer, and add tests covering its use with Postgres. Support for other database types will follow later (see #157, #158).

As part of this change, make `Hanami::Providers::DB#database_url` public API, and access it through `slice.container.providers[:db].source.database_url` (with `Dry::System::Provider#source` soon to be tagged as public API too).

This ensures that the canonical databases across an app can be identified via their database URL strings alone, without having to establish a full connection to the database. This means all the `db` CLI commands (not just `structure dump`) can operate more efficiently. It also makes them easier to test inside the repo, since we can avoid providing the live database servers for them to connect to.

Resolves #154.

### DIFF
--- a/lib/hanami/cli/commands/app/db/command.rb
+++ b/lib/hanami/cli/commands/app/db/command.rb
@@ -17,6 +17,17 @@ module Hanami
             option :app, required: false, type: :boolean, default: false, desc: "Use app database"
             option :slice, required: false, desc: "Use database for slice"
 
+            attr_reader :system_call
+
+            def initialize(
+              out:, err:,
+              system_call: SystemCall.new,
+              **opts
+            )
+              super(out: out, err: err, **opts)
+              @system_call = system_call
+            end
+
             private
 
             def databases(app: false, slice: nil)
@@ -30,24 +41,23 @@ module Hanami
             end
 
             def database_for_app
-              Utils::Database[app]
+              build_database(app)
             end
 
             def database_for_slice(slice)
               slice = inflector.underscore(Shellwords.shellescape(slice)).to_sym
 
-              Utils::Database[app.slices[slice]]
+              build_database(app.slices[slice])
             end
 
             def all_databases
               slices = [app] + app.slices.with_nested
 
               slices_by_database_url = slices.each_with_object({}) { |slice, hsh|
-                next unless slice.container.providers[:db]
+                provider = slice.container.providers[:db]
+                next unless provider
 
-                slice.prepare :db
-                database_url = slice["db.gateway"].connection.uri
-
+                database_url = provider.source.send(:database_url)
                 hsh[database_url] ||= []
                 hsh[database_url] << slice
               }
@@ -55,7 +65,7 @@ module Hanami
               databases = slices_by_database_url.each_with_object([]) { |(url, slices), arr|
                 slices_with_config = slices.select { _1.root.join("config", "db").directory? }
 
-                database = Utils::Database[slices_with_config.first || slices.first]
+                database = build_database(slices_with_config.first || slices.first)
 
                 warn_on_misconfigured_database database, slices_with_config
 
@@ -63,6 +73,10 @@ module Hanami
               }
 
               databases
+            end
+
+            def build_database(slice)
+              Utils::Database[slice, system_call: system_call]
             end
 
             def warn_on_misconfigured_database(database, slices)
--- a/lib/hanami/cli/commands/app/db/command.rb
+++ b/lib/hanami/cli/commands/app/db/command.rb
@@ -57,7 +57,7 @@ module Hanami
                 provider = slice.container.providers[:db]
                 next unless provider
 
-                database_url = provider.source.send(:database_url)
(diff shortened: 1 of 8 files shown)


lib/hanami/cli/commands/app/db/command.rb
```diff
@@ -30,32 +41,31 @@ def databases(app: false, slice: nil)
...
 
               slices_by_database_url = slices.each_with_object({}) { |slice, hsh|
-                next unless slice.container.providers.find_and_load_provider(:db)
-
-                slice.prepare :db
-                database_url = slice["db.gateway"].connection.uri
+                provider = slice.container.providers.find_and_load_provider(:db)
+                next unless provider
 
+                database_url = provider.source.send(:database_url)
```
timriley, line 60: Avoid unnecessary connections to the database by reaching inside the provider source and calling its `#database_url` method (which we presume to be present). Right now `#database_url` is a private method, but we can make it public before this PR merges.

In addition, this method of accessing a provider object itself is not proper public API within dry-system. I think we need to streamline it:

- Have `Dry::System::ProviderRegistrar#[]` invoke `find_and_load_provider` so callers don't need to worry about that step (and to ensure consistent behaviour for provider registrars that aren't finalised)
- Classify `Dry::System::ProviderRegistrar#[]` as `@api public`
- Classify `Dry::System::Provider#source` as `@api public`
- Possibly even consider having `Dry::System::Provider` delegate missing methods to its source for nicer ergonomics, though I think we can skip this for now.

Why have I needed to do this?

Without this change, we had to establish a database connection just to get a slice's database_url, which we already know from config alone! This is wasteful and pointless. Worse still, we would actually connect multiple times to the same database for an app with multiple slices using the same database URL.

Why do we need to call this `provider.source.database_url`, though? This is to make sure we're using the "source of truth" for this database URL, which is the provider source itself, since it contains the logic to determine this URL when it eventually does connect.

It also makes testing this CLI command much harder (i.e. forcing us to have running database servers).

@solnic @flash-gordon @alassek — I'd be interested in your thoughts on this kind of provider usage. This feels reasonable to me, but it does make providers potentially "richer" objects in regular usage. I think this is safe though, since you couldn't confidently call methods against a provider's source unless you were already confident about what it should be.

An alternative would be to find some other home for this database_url logic, e.g. a static method somewhere in hanami-db itself. For this to work, though, we'd still need to send config from the provider to that method, since a user may choose to override the database_url directly in their own provider code. This is why I decided to stick with the provider as the source of truth and just use that directly.

timriley:
I forgot to push some expanded tests before merging. FYI, they are here: https://github.com/hanami/cli/commit/cf0f88f3655da3e4d66d951edf9eb083a3589873

lib/hanami/cli/commands/app/db/command.rb
timriley: After a few days wait, I've gone ahead and merged this. I think it's a reasonable approach, and since this is all internal (i.e. no public API commitments), we can change it in the future if we need.

alassek:
@timriley I think this may indicate the need for an abstract DB Config object, which is the direction I went with `Persistence::Specification` encapsulating things like URL, extensions, migration paths and such.

Since we will undoubtedly need this anyway for hanami/hanami#1389 and hanami/hanami#1409, perhaps we should move the `database_url` source of truth there.

timriley:
@alassek Can you please share more of how this looked for you?

In terms of having an explicit config object, we'll already have this in the provider's own `config`.

In what we have laid out so far, this feels like the right place to have it, right next to the code that needs to consume that config.

Improve provider config is what I want to look at next. My first step was to allow provider sources to declare their own custom config classes (which is a dry-configurable feature). This would mean that `config` within the `db` provider can become a custom class that we can add our own API into, which I think would help for more fluent setup of ROM plugins and the like.

e.g.

```
Hanami.app.configure_provider :db do
  config.rom.plugin :sql, :some_plugin do |plugin|
    # plugin config goes here

    # this block can then later be evaluated in the context of the provider,
    # so it can access app resources if needed
  end
end
```

This class could also provider reader methods for getting the config back out in various structures that are useful to us.

If there's anything else you think we should be considering at this stage, I'd be very happy to hear it!

FWIW, I'm actually done with all the `db` CLI commands now, and I didn't need anything more than just `database_url` from the provider source.

lib/hanami/cli/commands/app/db/command.rb
flash-gordon: @timriley I cannot see any problem with your approach. Is setting a URL is the only way to configure a DB connection?

Issue #166, opened 2024-06-14
Include dry-operation gem and generate base operation class in new apps and slices


timriley:

timriley:
Fixed by https://github.com/hanami/cli/pull/171.

PR #171, opened 2024-06-15
Add dry-operation generators

cllns:
Addresses #166. This creates a base operation in the app, then one in each slice when they're generated. Also adds a `hanami generate operation` command

Two issues that still need to be addressed:
1. ~Generating a top-level operation (i.e. `add_book` not `admin.books.add`) has empty lines around the class due to trying to add modules for namespaces that don't exist. The newlines from the ERB template are still there. A simple way to handle it would be `nested_app_operation.erb` and `nested_slice_operation.erb` but maybe we can think of something better?~ I added different templates for top-level and nested operations to solve this. Open to other solutions, but this is simple.

2. ~I have an `xit` for doing `admin/books/add` because that should be the same as `admin.books.add` but it's rendering it as `class Admin::Books::Add` instead of nested modules. Feel free to remove it but I feel like we should support this syntax the same way as the period separated names.~ Fixed this. This allows `hanami generate operation admin/books/add` to work just like `hanami generate admin.books.add`. This will match the folder structure. The only thing here is that now we should add the same functionality to all our generators, which I'm happy to do in a separate PR after this one.

Also, @timriley  do we still do `x.x.x` as the version number to be changed at release time?

Happy to address any feedback, and also feel free to take this WIP and finish it as you wish @timriley, I don't want to hold you up.


spec/unit/hanami/cli/commands/app/generate/operation_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      subject.call(name: "add_book")
+
+      operation_file = <<~EXPECTED
+        # frozen_string_literal: true
+
+        module Test
+          module Operations
+            class AddBook < Test::Operation
+              def call(input)
+              end
```
timriley, line 30: I worry about `call(input)` as a default signature making it feel like a single positional param is the only way this `#call` method can work (which is certainly not the case). What do you think about making it just `def call`?

lib/hanami/cli/commands/app/generate/operation.rb
```diff
@@ -0,0 +1,50 @@
...
+        module Generate
+          # @since x.x.x
+          # @api private
+          class Operation < App::Command
+            argument :name, required: true, desc: "Operation name"
+            option :slice, required: false, desc: "Slice name"
+
+            example [
+              %(add_book               (MyApp::Operations::AddBook)),
+              %(add_book --slice=admin (Admin::Operations::AddBook)),
```
timriley, line 22: I wonder if we want to provide a different example here, e.g. `"books.create"` -> `(MyApp::Books::Create)`, either in addition or instead of this one.

The "operations layer" is where we really get into the "domain" of apps, and this is where I want to make sure we encourage the user to really take control of their app structure. So if they want to start organising by "concept" at this point instead of by file type, then I think we should encourage it.

What do you reckon?

lib/hanami/cli/generators/app/operation.rb
```diff
@@ -0,0 +1,75 @@
...
+          attr_reader :fs
+
+          attr_reader :inflector
+
+          def generate_for_slice(context, slice)
+            slice_directory = fs.join("slices", slice)
+            raise MissingSliceError.new(slice) unless fs.directory?(slice_directory)
+
+            if context.namespaces.any?
+              fs.mkdir(directory = fs.join(slice_directory, "operations", context.namespaces))
```
timriley, line 44: I don't think we should hardcode `"operations"` in the directory path here. Operations should be able to be created _anywhere_ within an app or slice.

(This comment also applies to all the other places we've hard-coded "operations" within this PR)

timriley (changes requested):
This is looking great, thanks for putting it together, @cllns!

Biggest piece of feedback from me is that I want us to allow users to generate operations into any place in their apps or slices.

In this way `generate operation` is just a more specialised version of the `generate component` command we've recently added. 

Have left some inline comments explaining more of my thinking here.

Would you be up for making that adjustment? Thank you! :heart:

cllns:
All great feedback, I agree with all of it. Will work to implement it. Still unsure about the exact pattern we want but I can get us closer to allowing operations anywhere, even if we tweak the UX later

lib/hanami/cli/commands/app/generate/operation.rb
cllns: Yea, I agree.

A very small thing but how do you feel about calling it "books.add" though, instead of "books.create". 

For a publishing project, `create` would make sense (making a new book project that didn't exist before), but for a bookshelf project, the physical book already exists, it's merely being `add`ed to the bookshelf system. I feel like it's a good domain-driven design convention we can incorporate. 

Plus it can help people avoid thinking that "create" has something to do with an action or repo, since both of those often have "create". We want to let people think in terms of their domain, not the implementation.

cllns:
Ok I implemented removing the "Operations" folder and namespace, so now it's completely generic.

I also made it (in my [latest commit](https://github.com/hanami/cli/pull/171/commits/a5bd2f30311293c24594d4c84bb0c1033dd055e3)) so that users **must** provide a namespace. I think this is fair, since `app/` (and `slices/SLICE_NAME`) only have folders and base classes in it. I feel like it's not a great spot for non-abstract classes. People can still move them there after generation if they want, but I don't think we should encourage it. WDYT?

Note that this is different from how component is currently generated, which lets people generate without the namespace. If we do this here, perhaps we should do it there too.

I worry the error message I wrote is too verbose, WDYT?

lib/hanami/cli/commands/app/generate/operation.rb
timriley: Yep, that's fine with me 👍🏼

cllns:
Ok, I got rid of the Error and now we have it as puts-ing a `"Recommendation:"` that's indented two spaces to visually distinguish it from the other output.

I added `Hanami::CLI::Files#recommend`, which is a slightly awkward place since it's just outputting, but it's related to file generation so I think it's a practical place for it.

I started to add skip_recommendations as CLI arg but it's awkward and ends up touching a lot of stuff, so maybe we can wait until people complain to build that out?

cllns:
I just integrated dry-operation into my app and I had to add `include Dry::Monads[:result]` in a couple spots. In my app it was `app/action.rb` and `spec/spec_helper.rb` (where I also had to `require "dry/monads"`) in order to be able to match on the result.

I was thinking we could just include it generally within the main App namespace by default, but the specs use inline syntax so that doesn't work due to nesting.

Where should we `include Dry::Monads[:result]` when an operation is generated? At the very least we need to output a message instructing the users to include it, but ideally we could just include it in a few key spots when generating the first operation so it would Just Work. We could also add it when generating a new app, but the M-word scares people sometimes.

cllns:
Also, should we add [ROM transaction support](https://github.com/dry-rb/dry-operation/pull/15) to the default Operation if hanami-db is loaded? Related to #180


I think it would be something like:
```
class MyApp
  class Operation < Dry::Operation
    include Dry::Operation::Extensions::ROM
    include Deps[rom: "db.rom"]
  end
end
```

lib/hanami/cli/files.rb
```diff
@@ -42,6 +42,12 @@ def chdir(path, &blk)
         super
       end
 
+      # @since x.x.x
+      # @api private
+      def recommend(message)
+        out.puts("  Recommendation: #{message}")
+      end
```
timriley, line 49: This as part of our `fs` object feels weird. Is there any reason we can't output this from inside our generator class? We could include `out:` as a dependency to that class.

Then it would just be:

```
fs.write # ...
out.puts("Add a namespace...")
```

lib/hanami/cli/generators/app/operation.rb
```diff
@@ -0,0 +1,77 @@
...
+            slice_directory = fs.join("slices", slice)
+            raise MissingSliceError.new(slice) unless fs.directory?(slice_directory)
+
+            if context.namespaces.any?
+              fs.mkdir(directory = fs.join(slice_directory, context.namespaces))
+              fs.write(fs.join(directory, "#{context.name}.rb"), t("nested_slice_operation.erb", context))
+            else
+              fs.mkdir(directory = fs.join(slice_directory))
+              fs.write(fs.join(directory, "#{context.name}.rb"), t("top_level_slice_operation.erb", context))
+              fs.recommend("Add a namespace to operation names, so they go into a folder within #{directory}/.")
```
timriley, line 49: What do you think about this for wording?

```suggestion
              fs.recommend(%(Generating a top-level operation. To generate into a directory, add a namespace: "my_namespace.#{context.name}))
```

That first sentence is there to explain why we're offering the recommendation in the first place. The second sentence is there to give a bit more guidance as to what the "recommended" args might look like.

timriley (commented):
Nearly there! Just a couple last thoughts. Thanks for bearing with me here, @cllns :)

lib/hanami/cli/files.rb
cllns: Yup, that makes sense! I did this

lib/hanami/cli/generators/app/operation.rb
cllns: Sweet, switched it to this. I added the indentation too so it's more noticeable.

timriley (approved):
@cllns Looks good! Let's add `@since 2.2.0` instead of `@since x.x.x`. After that's done, feel free to merge! Thank you!

cllns:
> I just integrated dry-operation into my app and I had to add `include Dry::Monads[:result]` in a couple spots. In my app it was `app/action.rb` and `spec/spec_helper.rb` (where I also had to `require "dry/monads"` and `config.include`) in order to be able to match on the result.
> 
> I was thinking we could just include it generally within the main App namespace by default, but the specs use inline syntax so that doesn't work due to nesting.
> 
> Where should we `include Dry::Monads[:result]` when an operation is generated? At the very least we need to output a message instructing the users to include it, but ideally we could just include it in a few key spots when generating the first operation so it would Just Work. We could also add it when generating a new app, but the M-word scares people sometimes.

I also added `include Dry::Monads[:result]` into the base app Action, so people can match on the result of operations in their actions. What do we think about this? If we do this, I can also make a PR to `hanami-rspec` to add the require and `config.include Dry::Monads[:result]` in the rspec support file, so testing operation just works.

lib/hanami/cli/generators/gem/app/action.erb
```diff
@@ -5,5 +5,6 @@ require "hanami/action"
 
 module <%= camelized_app_name %>
   class Action < Hanami::Action
+    include Dry::Monads[:result]
```
timriley, line 8:
```suggestion
    # For pattern matching on operation results
    include Dry::Monads[:result]
```

I like that you've included this!

I suggest we include a little explainer to make it clear why we're including this.

timriley: For consistency, can you please update the slice generator to include this too?

The difference should be that the slice generator should check to see if dry-monads is bundled before it includes this line. If the user chooses to remove dry-operation from their `Gemfile`, then dry-monads will disappear and the `include` would then result in an error.

timriley (approved):
Thanks again @cllns! This is looking good :)

> I also added include Dry::Monads[:result] into the base app Action, so people can match on the result of operations in their actions with just Success(value) instead of needing to write Dry::Monads::Result::Success in every spot they use a dry-operation. What do we think about this?

Yep, I think this is great!

I left two tiny pieces of feedback about this via inline comment. Once those are done, let's merge this!

> If we do this, I can also make a PR to hanami-rspec to add the require and config.include Dry::Monads[:result] in the rspec support file, so testing operations just works in the same way.

Yeah, that would be awesome!

lib/hanami/cli/generators/gem/app/action.erb
cllns: I did this for the slice, but the slice actions inherit from app/action.rb so it doesn't do anything. I kept it in since I might be missing something, but I think we can remove it?

I also had to add a line to require dry/monads

timriley: @cllns Ah yeah, sorry, that's my mistake. With slice actions inheriting from an app action, we don't need to put anything additional into the slice actions.

timriley (approved):
This is still looking great, thanks @cllns!

I just [noted here](https://github.com/hanami/cli/pull/171#discussion_r1666782867) that we indeed don't need to be touching slice actions — sorry for the runaround there.

Once that's sorted out again, this looks good to merge!

PR #193, opened 2024-07-09
Move constants used in generators

krzykamil:
Based on https://github.com/hanami/cli/pull/130#discussion_r1627648954 I moved the constants commonly used in all generators to a common place. This also required a simple change to one of the context classes, as it was using OFFSET that was an integer, instead of the one that is an alias to INDENTATION and work better with multiplication in the code that generates the spacing in end files.


cllns:
Thanks @krzykamil! I felt bad about copying them yet again when I did Operation generator, so I'm glad this finally got done :)

PR #195, opened 2024-07-10
Use dry-cli's :flag type instead of :boolean

cllns:
I saw a bunch of `--[no]-skip-view` etc, which doesn't make sense. I realized it's because we're using `type: :boolean` instead of `type: :flag`.

I went though each one and I think they can all be converted to flags without losing anything, so that's what I did here.

This **does not currently work** so should not be merged, but it should work once https://github.com/dry-rb/dry-cli/pull/117 is brought up-to-date and merged. I'll handle fixing that up tomorrow, and change this PR to a non-draft once that's merged.

### DIFF
--- a/lib/hanami/cli/commands/app/db/migrate.rb
+++ b/lib/hanami/cli/commands/app/db/migrate.rb
@@ -10,8 +10,8 @@ module Hanami
             desc "Migrates database"
 
             option :target, desc: "Target migration number", aliases: ["-t"]
-            option :dump, required: false, type: :flag, default: true,
-              desc: "Dump the database structure after migrating"
+            option :dump, required: false, type: :boolean, default: true,
+                          desc: "Dump the database structure after migrating"
 
             def call(target: nil, app: false, slice: nil, dump: true, command_exit: method(:exit), **)
               databases(app: app, slice: slice).each do |database|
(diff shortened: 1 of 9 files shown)


timriley:
Thanks for taking this on, @cllns! I had also noticed it and registered it as a bit odd, so I'm really glad we'll straighten this out before release!

lib/hanami/cli/commands/app/db/migrate.rb
```diff
@@ -10,7 +10,7 @@ class Migrate < DB::Command
             desc "Migrates database"
 
             option :target, desc: "Target migration number", aliases: ["-t"]
-            option :dump, required: false, type: :boolean, default: true,
+            option :dump, required: false, type: :flag, default: true,
```
timriley, line 13: I think we want this to be a `:boolean` still? Since this defaults to true, the thing that the user will be passing is actually `--no-dump`.

Can you double check we haven't mistakenly converted any other `:boolean` options with a default of `true`?

cllns: Good call, fixed and confirmed there are no others.

timriley:
@cllns Awesome! Is this ready to merge?

timriley:
I've popped the main branch for dry-cli into the `Gemfile`, even though the tests weren't failing before (which is another indication we need some higher-level integration tests).

Will merge this in now just to keep everything moving.

PR #198, opened 2024-07-12
Change component generator examples to domain-specific ones

cllns:
The component generator was created before the operation generator. It's potentially confusing to generate a component called "operation", since we have an operation generator.

~I don't think services is a terrific alternative (since it's often used in the same way as operation) so please let me know if anyone has any better ideas. Just wanted to open this so I don't forget~
I brainstormed some domain-specific components someone could want for a bookshelf app. Open to other suggestions (in particular ISBN should be an acronym, but it's not in the default inflector, so I think it's find to render it as `Isbn`, but it's not ideal)


timriley:
Thanks @cllns!

PR #199, opened 2024-07-13
Generate struct (with RubyFileGenerator, add RubyFileWriter)

cllns:
This adds a 'struct' generator, and some abstractions, a `Hanami::CLI::Commands::App::Generate::Command`, which handles the creation of a generator object, as well as calling the generator.

We also add a RubyFileWriter class (I don't love the name but it seems OK). This determines what gets passed to RubyFileGenerator, deciding on name of the class, the body, the modules it's nested in, the parent, etc. It's got 8 initializer kwarg params which is a bit much. It can be refactored later but for now it's an implementation detail so I think it's OK. We could break out some of them (relative_parent_class, extranamespace, body) into args on `call` but then we have to pass them around more, too. Happy to do that if we prefer, though.

Note that this PR is against the branch for #186.


lib/hanami/cli/commands/app/generate/command.rb
```diff
@@ -25,15 +25,17 @@ class Command < App::Command
...
-
-              super(fs: fs, inflector: inflector, **opts)
+              super
               @generator = generator_class.new(fs: fs, inflector: inflector, out: out)
             end
 
+            def generator_class
+              # Must be implemented by subclasses, with class that takes:
+              # fs:, inflector:, out:
+            end
```
timriley, line 37: Ahhhh, you did an equivalent to what I suggested in https://github.com/hanami/cli/pull/186, so we're deinfitely on the same page here 😄

lib/hanami/cli/commands/app/generate/struct.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require "dry/inflector"
+require "dry/files"
+require "shellwords"
+require_relative "../../../naming"
+require_relative "../../../errors"
```
timriley, line 7:
```suggestion
require "dry/inflector"
require "dry/files"
require "shellwords"
require_relative "../../../naming"
require_relative "../../../errors"
```

I don't think we need any of these requires here anymore! Feel free to take them out of this as well as the `Generate::Operation` command before merging 😄

lib/hanami/cli/generators/app/ruby_file_writer.rb
```diff
@@ -0,0 +1,149 @@
...
+          # @api private
+          def initialize(
+            fs:,
+            inflector:,
+            app_namespace:,
+            key:,
+            slice:,
+            relative_parent_class:,
+            extra_namespace: nil,
+            body: []
```
timriley, line 25: Like you mentioned in the description, this is an internal class only so we can tolerate some slightly less ergonomic API design, but I think a refactor to allow a FileWriter instance to be initialised once and then used to write multiple files might be nice.

So this would probably pull the `fs:` and `inflector:` params out into `#initialize`, with the rest going into some method that writes an actual file.

I think this'd be worth playing with sometime, but it's not blocking and I'm definitely keen to get our full range of CLI commands ready for beta1 on Tuesday. So happy if this comes in a follow-up change of some kind :)

spec/unit/hanami/cli/commands/app/generate/struct_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  let(:out) { StringIO.new }
+  let(:fs) { Hanami::CLI::Files.new(memory: true, out: out) }
+  let(:inflector) { Dry::Inflector.new }
+  let(:app) { Hanami.app.namespace }
+  let(:dir) { inflector.underscore(app) }
+
+  def output
+    out.rewind && out.read.chomp
+  end
```
timriley, line 14:
```suggestion
  def output
    out.string.chomp
  end
```

I realised there's a way nicer way to do this that avoids mutating the IO object itself.

spec/unit/hanami/cli/commands/app/generate/struct_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        end
+      EXPECTED
+
+      expect(fs.read("app/structs/book/published_book.rb")).to eq(struct_file)
+      expect(output).to include("Created app/structs/book/published_book.rb")
+    end
+  end
+
+  context "generating for a slice" do
+    it "generates a struct in a top-level namespace, with recommendation" do
```
timriley, line 77:
```suggestion
    it "generates a struct in a top-level namespace" do
```

I think "with recommendation" might just be a hangover from the operation generator tests?

lib/hanami/cli/generators/app/struct.rb
```diff
@@ -0,0 +1,44 @@
...
+require_relative "../constants"
+require_relative "../../errors"
+
+module Hanami
+  module CLI
+    module Generators
+      module App
+        # @since 2.2.0
+        # @api private
+        class Struct
```
timriley, line 14: These new generators are _so tidy_ 😍

timriley (approved):
Looking great, let's get this in!

A couple of tidy-up notes, feel free to take care of those before merging, no further review necessary 👍🏼

lib/hanami/cli/commands/app/generate/struct.rb
cllns: Oh right, great! Gone

spec/unit/hanami/cli/commands/app/generate/struct_spec.rb
cllns: Sweet. Did this for struct and operation, and made a note to self to fix these for the other generators when I switch them over to using this RubyFileGenerator/Writer

lib/hanami/cli/generators/app/ruby_file_writer.rb
cllns: Made a note for myself to explore this later

spec/unit/hanami/cli/commands/app/generate/struct_spec.rb
cllns: Busted! Fixed this

PR #200, opened 2024-07-13
Generate repo

### LINKED ISSUE #189: `generate repo` command

cllns:
Builds on work done in #199. This is where those refactors pay off: this almost trivial PR to add a repo generator. 

I decided to add some logic for ~pluralization~ (update:) singularization:
`hanami g repo book` => BookRepo
`hanami g repo books` => BookRepo
`hanami g repo books_repo` => **BooksRepo**
`hanami g repo book_repo` => BookRepo

The idea here is that we want to guide people to naming things with our convention, but if they end their name with _repo then we assume they're doing it intentionally and want that exact name, without any singularization.

Closes #189


lib/hanami/cli/commands/app/generate/repo.rb
```diff
@@ -0,0 +1,49 @@
...
+              Generators::App::Repo
+            end
+
+            # @since 2.2.0
+            # @api private
+            def call(name:, slice: nil, **opts)
+              normalized_name = if name.end_with?("_repo")
+                                  name
+                                else
+                                  "#{inflector.pluralize(name)}_repo"
```
timriley, line 38: I love the idea of us providing the combination of convenience _and_ guard rails through our arg handling!

```suggestion
                                  "#{inflector.singularize(name)}_repo"
```

Just one adjustment though: repos are actually named singularly, so `BookRepo`, not `BooksRepo`.

lib/hanami/cli/commands/app/generate/repo.rb
```diff
@@ -0,0 +1,49 @@
...
+                                end
+              slice = inflector.underscore(Shellwords.shellescape(slice)) if slice
```
timriley, line 40: TBH, I don't think we need to shellescape this slice arg at all. We're checking for the presence of the slice dir via our generators anyway, so that's validation enough for the slice name.

In https://github.com/hanami/cli/pull/201 I removed the shellescaping, and I reckon we could just do the same here too.

timriley (approved):
This all looks great!

If you could take care of switching from pluralised to singularised names, then we can get this in!

lib/hanami/cli/commands/app/generate/repo.rb
cllns: Oops, duh! Fixed

PR #201, opened 2024-07-13
Add `generate migration` command and generator

### LINKED ISSUE #179: Generate migrations with `generate migration`
For implementation, see the `db create_migration` command that I removed in https://github.com/hanami/cli/pull/178. We can probably reuse a lot of its code. (Although we shouldn't feel beholden to it; if we can see a better approach, let's feel free to do that!)

timriley:
Resolves #179 

As I was building this, I realised we have quite a bit of double handling across our `generate <foo>` CLI commands and the actual generators they call. To help centralise the logic in this case, I moved the generator-specific validation checks into the generator itself.

This keeps the command class much thinner and mostly just responsible for "writing".

We could consider refactoring our other commands/generators like this in the future, but it's not urgent.

### DIFF
--- a/lib/hanami/cli/commands/app/generate/migration.rb
+++ b/lib/hanami/cli/commands/app/generate/migration.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module Hanami
+  module CLI
+    module Commands
+      module App
+        module Generate
+          # @since 2.2.0
+          # @api private
+          class Migration < App::Command
+            argument :name, required: true, desc: "Migration name"
+            option :slice, required: false, desc: "Slice name"
+
+            example [
+              %(create_posts),
+              %(add_published_at_to_posts),
+              %(create_users --slice=admin),
+            ]
+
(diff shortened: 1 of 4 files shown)


timriley:
I realise I forgot to `register` the `"generate migration"` command (that we have nothing codified to remind me of this is a downside of our current testing approach we should remedy sometime!)

I/we should make sure we add it before merging.

cllns:
> I realise I forgot to `register` the `"generate migration"` command (that we have nothing codified to remind me of this is a downside of our current testing approach we should remedy sometime!)

Yea, maybe we could do an integration spec for each of them, just check that the right text is outputted and that a file is created where we expect. Maybe for the examples too?

cllns:
I started trying to convert our RubyFileWriter to support this case (a big block instead of a class definition in a file) but ultimately decided to keep the existing strategy, I just refactored it a little. I can't add you as a reviewer since you made the PR, but let me know what you think @timriley

PR #206, opened 2024-07-14
Generate relation

### LINKED ISSUE #188: `generate relation` command

cllns:
Resolves #188 

Pretty straight-forward. I made the decision to pluralize the name so `generate relation book` generates a `Books` relation.

And there's no way to allow a singularized relation, which is fine I think because that'd likely conflict with the associated Struct.

But I did make it so that if people choose to use a `_relation` suffix (which we don't recommend) we allow them to do that and keep their core name as what they specified. The reason is that we don't want to make `BookRelations`, that'd be very wrong.

I also cleaned up/expanded some specs for struct/repo/operation.

### DIFF
--- a/lib/hanami/cli/commands/app.rb
+++ b/lib/hanami/cli/commands/app.rb
@@ -45,6 +45,7 @@ module Hanami
               prefix.register "component", Generate::Component
               prefix.register "operation", Generate::Operation
               prefix.register "part", Generate::Part
+              prefix.register "relation", Generate::Relation
               prefix.register "repo", Generate::Repo
               prefix.register "slice", Generate::Slice
               prefix.register "struct", Generate::Struct
--- a/lib/hanami/cli/commands/app/generate/relation.rb
+++ b/lib/hanami/cli/commands/app/generate/relation.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Hanami
+  module CLI
+    module Commands
+      module App
+        module Generate
+          # @since 2.2.0
+          # @api private
+          class Relation < Command
+            argument :name, required: true, desc: "Relation name"
+            option :slice, required: false, desc: "Slice name"
+
+            example [
+              %(books               (MyApp::Relation::Book)),
+              %(books/drafts        (MyApp::Relations::Books::Drafts)),
+              %(books --slice=admin (Admin::Relations::Books)),
+            ]
+
+            # @since 2.2.0
+            # @api private
+            def generator_class
+              Generators::App::Relation
+            end
+
+            # @since 2.2.0
+            # @api private
+            def call(name:, slice: nil, **opts)
+              normalized_name = if name.end_with?("_relation")
+                                  name
+                                else
+                                  inflector.pluralize(name)
+                                end
+
+              super(name: normalized_name, slice: slice, **opts)
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/hanami/cli/generators/app/operation.rb
+++ b/lib/hanami/cli/generators/app/operation.rb
@@ -1,9 +1,6 @@
 # frozen_string_literal: true
 
-require "erb"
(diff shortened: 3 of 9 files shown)


timriley:
Hey @cllns! Thanks for getting this up! I've just pushed one commit with a couple changes:

- Don't attempt any relation name normalization. Relation names are sensitive, in that they typically match the name of their underlying data source. And because real-world databases are funky, I can imagine scenarios where people are using non-typical names. Given this, I don't think we should respect whatever name is passed into our generator.
- I've also made it so these generate the necessary `schema :relation_name, infer: true` in the class body. Have I don't it correctly? For deeply namespaced relations, I'm just taking the final segment of the key, which feels reasonable.
- For what it's worth, we don't actually _register_ deeply nested relations when booting our db layer at the moment, but this should be supported, so I made a ticket for it: https://github.com/hanami/hanami/issues/1441. I'm fine for us to be able to generate nested relations as of this upcoming beta, though; I don't think people will _really_ do this much, and we'll fix the other side soon enough.

If you're fine with all of these, I'd be happy for you to merge!

timriley (approved):
All looks great! I pushed a change so I left a comment for you to consider, if you have the time. Otherwise I'll merge this in about 22 hours anyway :)

cllns:
> Hey @cllns! Thanks for getting this up! I've just pushed one commit with a couple changes:
> 
>  * Don't attempt any relation name normalization. Relation names are sensitive, in that they typically match the name of their underlying data source. And because real-world databases are funky, I can imagine scenarios where people are using non-typical names. Given this, I don't think we should respect whatever name is passed into our generator.

Isn't that use-case handled flexibly by the schema definition? e.g. a `Books` relation could have `schema :legacy_book_table` and work perfectly fine? This is a benefit of the Relation being an abstraction on top of the dataset (i.e. DB table)

We could pluralize by default and add a `--force-name` option, too, but that'll introduce a convention that users might expect from other generators so we should tread lightly IMO. I'd love to have that eventually but it's not necessary soon.

If we don't pluralize then how do you feel about us adding a printed recommendation that we suggest making relation names plural? 



> * I've also made it so these generate the necessary `schema :relation_name, infer: true` in the class body. Have I don't it correctly? For deeply namespaced relations, I'm just taking the final segment of the key, which feels reasonable.

Ah yes, of course. I think the final segment as the generated default makes sense.

> * For what it's worth, we don't actually _register_ deeply nested relations when booting our db layer at the moment, but this should be supported, so I made a ticket for it: [Register nested relations hanami#1441](https://github.com/hanami/hanami/issues/1441). I'm fine for us to be able to generate nested relations as of this upcoming beta, though; I don't think people will _really_ do this much, and we'll fix the other side soon enough.
Ah, sweet!

timriley:
> Isn't that use-case handled flexibly by the schema definition? e.g. a Books relation could have schema :legacy_book_table and work perfectly fine? This is a benefit of the Relation being an abstraction on top of the dataset (i.e. DB table)

So it turns out that a relation's class name really not that important overall; it's actually the relation's _schema name_ that determines the name of its registration within the ROM container.

For example:

```
class WhatsUpDog < Hanami::DB::Relation
  schema :posts, infer: true
end
```

... will be registered as `rom.relations[:posts]`.

And:

```
class HelloYouFoolILoveYou < Hanami::DB::Relation
  schema :posts, as :joyrides
end
```

... will be registered as `rom.relations[:joyrides]`.

This is a potentially confusing quirk!

I think for Hanami's relations we ideally want to make the class name at least closely resemble the eventual registration name. So I think we can consider a pluralisation convention (the points you raised in the rest of your comment are good ones) for a post-2.2.0.beta1 version of this command.

Let's just get the basic one merged now, with no opinion on names, since I want our getting started guides to work for beta1 (they mention `generate relation`). This will buy us the time to properly consider the pluralisation etc. and lock in our preferred behaviour before the next release.

PR #208, opened 2024-07-16
Fix redundant name arg in generate commands

timriley:
Our recently updated `generate` subcommands are using a new `Commands::App::Generate::Command` base class, which had `argument :name` defined.

However, many of these subclasses _also_ had their own `argument :name` defined too, which led to these commands expecting _two_ name arguments to be given, which is incorrect.

I've fixed this by removing the `argument :name` from the base class, since having this in the subclasses allows them to provide their own specific `desc:` help text. After this change, commands like `generate relation create_posts` work again.


timriley:
cc @cllns

cllns:
Ah, sorry about that!

PR #218, opened 2024-07-27
Disable auto registration for base classes

cllns:
Cleanup for #180 and #197. 

We don't want the base classes to be auto_registered in the app's container, since they're meant to be used as parent classes, not used as instances directly.

### DIFF
--- a/lib/hanami/cli/generators/app/slice/relation.erb
+++ b/lib/hanami/cli/generators/app/slice/relation.erb
@@ -1,3 +1,4 @@
+# auto_register: false
 # frozen_string_literal: true
 
 module <%= camelized_slice_name %>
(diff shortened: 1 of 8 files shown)


timriley:
Thanks for being conscientious and looping back to this, @cllns!

However, I think we're already in the clear here: all of the files you touch here are generated into `db/` directories, and since `db` [is in our default `no_auto_register_paths`](https://github.com/hanami/hanami/blob/32802d69aa9bdc61086751b2e0405dfe3913a542/lib/hanami/config.rb#L114), nothing in this directory will be auto-registered anyway.

And I don't think we should put these directives on files that are already excluded courtesy of their location, _especially_ if it's not likely those files will ever move anywhere, since the unnecessary inclusion of the directive may confuse our users.

So I think we can close this one 😄

cllns:
Oh sweet, I missed that! Thanks

PR #220, opened 2024-07-29
Provide sample config/db/seeds.rb to guide people

francois:
Without the file, developers are left wondering where the seed should
live, and what should go in it. This guides the developer for the next
step.

### DIFF
--- a/lib/hanami/cli/commands/app/db/seed.rb
+++ b/lib/hanami/cli/commands/app/db/seed.rb
@@ -15,11 +15,13 @@ module Hanami
             def call(app: false, slice: nil, **)
               databases(app: app, slice: slice).each do |database|
                 seeds_path = database.slice.root.join(SEEDS_PATH)
-                next unless seeds_path.file?
-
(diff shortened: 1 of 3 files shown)


francois:
I didn't see any specs to check the contents of a new application. I did run `rm -rf bookshelf && bundle exec exe/hanami new bookshelf` and it correctly created the seed file.

I'd like to mention: the file is called `seeds.rb`, but the command is `hanami db seed`. Rails uses the same terminology, so maybe it's not a big surprise.

If the app has slices, I imagine we'd want to add seeds to the slices? This isn't taken into account in this PR, but happy to add, with some guidance. The guide needs some information about seeds as well. There are currently no database sections in the guides.

timriley:
Hi @francois! Thanks very much for putting this together. I had also made some recent changes in this direction, but your PR here made included aspects that I hadn't, such as more examples in the seeds file, and printing notices if the expected seeds files could not be found.

I incorporated these in #256 and included you as a co-committer. Thank you again! (And sorry for the delay)

Issue #202, opened 2024-07-14
Use slice.source_path in RubyFileWriter


cllns:
> If we can pass an actual slice instance into RubyFileWriter#directory, we could use our new `Slice#source_path` to make this simpler:
>
>```
>base = slice.source_path
>```
>
> This returns the appropriate value for both the app as well as slices, so it means we no longer need `if app else slice` conditionals everywhere, which has been a bother for the longest time :)
> 
>(Again, this would be another great refactor for post-merge, I don't want to hold anything up!)
>
_Originally posted by @timriley in https://github.com/hanami/cli/pull/186#discussion_r1676984531_

cllns:
When doing this, might be worth trying this other refactor for RubyFileWriter

> Like you mentioned in the description, this is an internal class only so we can tolerate some slightly less ergonomic API design, but I think a refactor to allow a FileWriter instance to be initialised once and then used to write multiple files might be nice.
>
>So this would probably pull the `fs:` and `inflector:` params out into `#initialize`, with the rest going into some method that writes an actual file.
>
>I think this'd be worth playing with sometime, but it's not blocking and I'm definitely keen to get our full range of CLI commands ready for beta1 on Tuesday. So happy if this comes in a follow-up change of some kind :)

_Originally posted by @timriley in https://github.com/hanami/cli/pull/199#discussion_r1676987192_

cllns:
Closed in #222 (though I actually didn't end up doing what this issue says)

PR #222, opened 2024-08-01
Refactor RubyFileWriter

cllns:
Addresses #202

Although, I ended up not using `slice.source_path`. I had an implementation working that just took a `Hanami::Slice` and used `.namespace` and `.source_path` but it proved hard to test. I couldn't figure out how to register a slice into the `Test::App`. 

Instead, the RubyFileWriter takes `namespace` and `base_path`, and we do the mapping in `Generate::Command`.

Happy to change this over if we can figure out a simple way to test that, to simplify the code, but I also feel like it's reasonable that a "File Writer" would require a namespace and a path as separate args, even though they're coupled in our apps.

I also changed the RubyFileWriter implementation so that `call` takes the file-specific args, instead of no args for `call` and all of them being passed to the initializer. I implemented this via helper class, as passing around all the variables that need to be passed around proved to make the code very messy.

I want to handle this refactoring before using RubyFileWriter in other generators.

### DIFF
--- a/lib/hanami/cli/commands/app/generate/repo.rb
+++ b/lib/hanami/cli/commands/app/generate/repo.rb
@@ -30,13 +30,9 @@ module Hanami
 
             # @since 2.2.0
             # @api private
-            def call(name:, slice: nil, **opts)
-              normalized_name = if name.end_with?("_repo")
-                                  name
-                                else
-                                  "#{inflector.singularize(name)}_repo"
-                                end
-              super(name: normalized_name, slice: slice, **opts)
+            def call(name:, **opts)
+              name = "#{inflector.singularize(name)}_repo" unless name.end_with?("_repo")
+              super
             end
           end
         end
--- a/lib/hanami/cli/generators/app/migration.rb
+++ b/lib/hanami/cli/generators/app/migration.rb
@@ -17,7 +17,7 @@ module Hanami
 
           # @since 2.2.0
           # @api private
-          def call(_app_namespace, name, slice, **_opts)
+          def call(_namespace, name, slice, **_opts)
             normalized_name = inflector.underscore(name)
             ensure_valid_name(normalized_name)
 
--- a/lib/hanami/cli/generators/app/migration.rb
+++ b/lib/hanami/cli/generators/app/migration.rb
@@ -17,8 +17,7 @@ module Hanami
 
           # @since 2.2.0
           # @api private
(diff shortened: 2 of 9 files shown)


lib/hanami/cli/commands/app/generate/repo.rb
```diff
@@ -30,13 +30,9 @@ def generator_class
...
             # @api private
-            def call(name:, slice: nil, **opts)
-              normalized_name = if name.end_with?("_repo")
-                                  name
-                                else
-                                  "#{inflector.singularize(name)}_repo"
-                                end
-              super(name: normalized_name, slice: slice, **opts)
+            def call(name:, **opts)
+              name = "#{inflector.singularize(name)}_repo" unless name.end_with?("_repo")
```
timriley, line 34: Just checking: by reassigning `name` here, that new value will still be passed up to the superclass method via the no-args version of `super` below?

lib/hanami/cli/generators/app/migration.rb
```diff
@@ -17,17 +17,13 @@ def initialize(fs:, inflector:, out: $stdout)
 
           # @since 2.2.0
           # @api private
-          def call(_app_namespace, name, slice, **_opts)
-            normalized_name = inflector.underscore(name)
-            ensure_valid_name(normalized_name)
+          def call(namespace:, key:, base_path:, **_opts)
+            _namespace = namespace
```
timriley, line 21: What's the reason for this extra assignment? It doesn't look like we use `_namespace` anywhere?

timriley (approved):
Thanks @cllns! These changes all look good to me 👍🏼 

Particularly in support for the adjustment to the `#initialize`/`#call` args for the `FileWriter`. This is much more consistent with how we design classes in general.

I left you a couple of questions, but they're minor, so happy for you to merge when ready.

lib/hanami/cli/commands/app/generate/repo.rb
cllns: Yup! Which I wasn't expecting. I wrote it explicitly as `super(name: name, **opts)`, but my Rubocop autocorrect got rid of it since it's equivalent to `super`.

lib/hanami/cli/generators/app/migration.rb
cllns: Ah, it was to appease Rubocop, but clearly the better move is to just remove the `namespace:` arg altogether, which is what I just committed.

Issue #209, opened 2024-07-16
Generators should remove the .keep file in a directory that they are populating


timriley:

cllns:
`Hanami::CLI::Files#write` should be a good place to implement this

PR #223, opened 2024-08-07
Deleting .keep files using the generators

kyleplump:
closes [#209 ](https://github.com/hanami/cli/issues/209)

I took @cllns 's advice and implemented this in `Hanami::CLI::Files#write`

Had to take care to skip this behavior when running `hanami new`, or else we'd never "keep" ( 😏 ) the files to begin with!  Let me know if there are any suggestions on improvements

Thanks!


cllns:
#224 is the successor to this PR

PR #224, opened 2024-08-07
Deleting .keep files using the generators

kyleplump:
(re-opened with correct commit history)

closes [#209 ](https://github.com/hanami/cli/issues/209)

I took @cllns  's advice and implemented this in `Hanami::CLI::Files#write`

Had to take care to skip this behavior when running hanami new, or else we'd never "keep" ( 😏 ) the files to begin with! Let me know if there are any suggestions on improvements

Thanks!

### DIFF
--- a/lib/hanami/cli/files.rb
+++ b/lib/hanami/cli/files.rb
@@ -19,15 +19,18 @@ module Hanami
       def write(path, *content)
         already_exists = exist?(path)
 
-        # delete .keep files unless running `hanami new`
-        base = path.split('/')[0]
-        keepfile = Dir.glob('**/.keep', base:)
+        if !directory?(path)
+          file = Pathname.new(path)
 
-        if caller_locations(1,1)[0].label != 'generate_app' && keepfile.any? && exist?("#{base}/#{keepfile[0]}")
-          delete("#{base}/#{keepfile[0]}")
+          file.ascend do |part|
+            potential_keepfile = (part.dirname + '.keep').to_path
+            delete(potential_keepfile) if exist?(potential_keepfile)
+          end
+
         end
 
         super
+
         if already_exists
           updated(path)
         else
--- a/lib/hanami/cli/files.rb
+++ b/lib/hanami/cli/files.rb
@@ -18,6 +18,15 @@ module Hanami
       # @api private
       def write(path, *content)
         already_exists = exist?(path)
+
+        # delete .keep files unless running `hanami new`
+        base = path.split('/')[0]
+        keepfile = Dir.glob('**/.keep', base:)
+
+        if caller_locations(1,1)[0].label != 'generate_app' && keepfile.any? && exist?("#{base}/#{keepfile[0]}")
+          delete("#{base}/#{keepfile[0]}")
+        end
+
         super
         if already_exists
           updated(path)
--- a/lib/hanami/cli/files.rb
+++ b/lib/hanami/cli/files.rb
@@ -19,18 +19,9 @@ module Hanami
       def write(path, *content)
         already_exists = exist?(path)
 
-        if !directory?(path)
-          file = Pathname.new(path)
-          file = file.relative_path_from(Pathname.new('/')) unless file.relative?
-
-          file.ascend do |part|
-            potential_keepfile = (part.dirname + '.keep').to_path
-            delete(potential_keepfile) if exist?(potential_keepfile)
-          end
-
-        end
-
         super
+
+        delete_keepfiles(path) unless already_exists
 
         if already_exists
           updated(path)
@@ -58,6 +49,24 @@ module Hanami
       private
 
       attr_reader :out
+
+      # Removes .keep files in any directories leading up to the given path.
+      #
+      # Does not attempt to remove `.keep` files in the following scenarios:
+      #   - When the given path is a `.keep` file itself.
+      #   - When the given path is absolute, since ascending up this path may lead to removal of
+      #     files outside the Hanami project directory.
+      def delete_keepfiles(path)
+        path = Pathname(path)
+
+        return if path.absolute?
+        return if path.relative_path_from(path.dirname).to_s == ".keep"
+
+        path.dirname.ascend do |part|
+          keepfile = (part + ".keep").to_path
+          delete(keepfile) if exist?(keepfile)
+        end
+      end
 
       def updated(path)
         out.puts "Updated #{path}"
--- a/lib/hanami/cli/files.rb
+++ b/lib/hanami/cli/files.rb
@@ -21,6 +21,7 @@ module Hanami
 
         if !directory?(path)
           file = Pathname.new(path)
+          file = file.relative_path_from(Pathname.new('/')) unless file.relative?
 
           file.ascend do |part|
             potential_keepfile = (part.dirname + '.keep').to_path
--- a/lib/hanami/cli/command.rb
+++ b/lib/hanami/cli/command.rb
@@ -23,7 +23,7 @@ module Hanami
       def self.new(
         out: $stdout,
         err: $stderr,
-        fs: Hanami::CLI::Files.new,
+        fs: Hanami::CLI::Files.new(out: out),
         inflector: Dry::Inflector.new,
         **opts
       )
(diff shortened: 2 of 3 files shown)


cllns:
Thanks for this @kyleplump! 

Can we take a different approach though? I think we can come up with somethin that avoids using `Kernel#caller_locations`. 

Perhaps we can just check the name of the file being written: if it has a name different than `.keep` then we delete `.keep` in the same folder?

kyleplump:
thanks for the feedback @cllns !

No problem on the `Kernel#caller_locations`.  But just so I understand, could you explain a little why you were hesitant to use it?  One thing i've noticed during implementing was that it pretty tightly coupled us to the fact that 'generate_app' will always be called that (dreaded string compare). but curious to hear your thoughts on this

you mean something like so?:
```
 if !directory?(path) && path.split('/').last != '.keep'
        base = path.split('/')[0]
        keepfile = Dir.glob('**/.keep', base:)
        delete("#{base}/#{keepfile[0]}") if keepfile.any? && exist?("#{base}/#{keepfile[0]}")
 end
```

im using glob because the path includes the subfolder (e.g. its never `app/actions` its always `app/actions/action_name/whatever.rb`).  is it safe to make an assumption that generators will always be namespaced one level down?  if so, instead of using glob i can just back up one directory and delete `.keep` there

the above code works for generators, but still runs during the initial `hanami new ...` call.  i cannot seem to find any 'state' or 'config' or way to tell in what context the `#write` action is being called, and dont really want to overload the function.  any ideas?

cllns:
> thanks for the feedback @cllns !
> 
> No problem on the `Kernel#caller_locations`. But just so I understand, could you explain a little why you were hesitant to use it? One thing i've noticed during implementing was that it pretty tightly coupled us to the fact that 'generate_app' will always be called that (dreaded string compare). but curious to hear your thoughts on this

My reasoning is that we should avoid using metaprogramming where possible. Metaprogramming is a powerful set of features and we make use of them in Hanami when needed, but only to provide a better end-user experience (as far as I know). It's harder to reason about, and like you said, it ties this code to the name of the method.

> 
> you mean something like so?:
> 
> ```
>  if !directory?(path) && path.split('/').last != '.keep'
>         base = path.split('/')[0]
>         keepfile = Dir.glob('**/.keep', base:)
>         delete("#{base}/#{keepfile[0]}") if keepfile.any? && exist?("#{base}/#{keepfile[0]}")
>  end
> ```
> 
> im using glob because the path includes the subfolder (e.g. its never `app/actions` its always `app/actions/action_name/whatever.rb`). is it safe to make an assumption that generators will always be namespaced one level down? if so, instead of using glob i can just back up one directory and delete `.keep` there
> 
> the above code works for generators, but still runs during the initial `hanami new ...` call. i cannot seem to find any 'state' or 'config' or way to tell in what context the `#write` action is being called, and dont really want to overload the function. any ideas?

Yea, something like that. I recommend using [`Pathname`](https://rubyapi.org/3.3/o/pathname). You can use `Pathname#split` to separate the name of the file from the folder it's in, and `Pathname#exist?` to see if it exists.

I see the point about deeper nested keep files being deleted too, but you can use `Pathname#glob` to get those.

Why would `hanami new` need any special behavior? If the file being created is a `.keep` file then it won't be immediately deleted. Doesn't that address the problem of not wanting to delete new `.keep` files?

kyleplump:
> My reasoning is that we should avoid using metaprogramming where possible. Metaprogramming is a powerful set of features and we make use of them in Hanami when needed, but only to provide a better end-user experience (as far as I know). It's harder to reason about, and like you said, it ties this code to the name of the method.

This makes a lot of sense, thank you for the explanation!

> I see the point about deeper nested keep files being deleted too, but you can use Pathname#glob to get those.

> Why would hanami new need any special behavior? If the file being created is a .keep file then it won't be immediately deleted. Doesn't that address the problem of not wanting to delete new .keep files?

i guess i'm still a bit confused here.  i definitely like the simplicity of your solution, but we don't write files to the same folder as a `.keep` files because its always nested under a folder  named after the 'generated thing'.  this is why i reached for using `glob`.  though, in using `glob`, i was inadvertently deleting the files when running the `new` command.   

for example these two lines in `App#generate_app`:

```
fs.write("app/actions/.keep", t("keep.erb", context))
fs.write("app/action.rb", t("action.erb", context))
```

works fine on the first `.write`, but searching for `.keep` using glob on the second like will remove the keep file.  this is the root of my conundrum here.  we could _not_ use glob, but then we're in the same position as before of tightly coupling our logic to file structure.  lmk if this makes sense or i'm way off base and not understanding in the first place

appreciate all your explanations / walking me through this :)

timriley:
@kyleplump I think the glob pattern you've been using is too greedy. When writing a new file (for example, `a/b/c.rb`), we only want to remove the following `.keep` files:

- in the very same directory as the file itself, e.g. `a/b/.keep`
- or in any parent directory, e.g. `a/.keep`

We do _not_ want to delete `.keep` files in deeper directories. So in this example, if there was an `a/b/nested/.keep` file, then writing `a/b/c.rb` should _not_ delete that file.

You glob of `Dir.glob('**/.keep', base:)` will look for nested files because of the `**`, I think. If we can restrict it to those two scenarios above, then all the file writing in `hanami new`—including the writing of the `.keep` files themselves-should work just as we expect, I think.

kyleplump:
thank you for the direction @timriley!

i've committed another attempt (still using `Pathname`, @cllns has converted me), and you're right - it seems traversing up the file path maintains the behavior of `hanami new` vs. the glob

let me know - thanks!

lib/hanami/cli/files.rb
```diff
@@ -18,7 +18,19 @@ def initialize(out: $stdout, **args)
       # @api private
       def write(path, *content)
         already_exists = exist?(path)
+
+        if !directory?(path)
+          file = Pathname.new(path)
+
+          file.ascend do |part|
```
timriley, line 25: This is all looking nice and tidy, thank you for continuing to iterate on this, @kyleplump!

One last thing I think we need to handle: I don't think we want to ascend all the to the root directory (`/`) checking for `.keep` files. Ideally we only ascend to the root of the Hanami project, and then stop there. Is this something you might be able to tackle?

Thanks!

timriley: Also, TIL `Pathname#ascend`! It is truly the class that keeps on giving 😍

kyleplump: @timriley thanks for the feedback!

in practice idk if we'd actually ascend up to the root directory because the given `path` was always a relative path (at least in my testing), but i added a new line to redefine the variable to be relative in case we ever found ourselves in this scenario (using our new workhorse, `Pathname` 😍 ) - let me know if this is what you were thinking.  thanks!

timriley: Thanks for taking on my feedback, @kyleplump! I had a think about this further and in the code I just pushed up (more on this in another comment), I think we can most safely handle this simply by bailing on attempting to delete .keep files if the given path is absolute. That way we're protected from ever going above the Dir.cwd or the Hanami app root.

And tbh, in actual practice, I don't _think_ we ever pass absolute files anyway, so this is fine.

timriley:
@kyleplump Thanks for your patience on this PR! I've just had a proper play around locally, made a few adjustments, and I think we're good to go now.

Here's what I did:

- Extracted the new logic into its own `delete_keepfiles` private method, which gives us the chance to early return in the exceptional cases, making for clearer code flow.
- Skip keepfile deletion if the path we're given is absolute. This is how we protect ourselves from accidentally removing files outside the Hanami project root. And like I mentioned in [my reply to you](https://github.com/hanami/cli/pull/224#discussion_r1730307475) above, I've confirmed from running our tests that we pass relative paths 100% of the time anyway.
- Skip keepfile deletion if the file we're creating is a `.keep` file itself. This means we don't disrupt the logic of `hanami new`.
- Only call `delete_keepfiles` when we're creating new files.

I also updated one of our generator tests to use a real app (instead of a dummy app and an in-memory file system) and add a test that a `.keep` file is appropriately deleted.

I'd love it if you could check it out again and see if it all still looks good to you. Thanks!

kyleplump:
hi @timriley! your thoughts make sense to me, and the code looks nice and clean. thanks for all of your guidance on this!

Issue #196, opened 2024-07-10
Add `--skip-route` option to `generate action` command


timriley:
If this is given, generating an action wouldn't attempt to update the routes file to match.

We could use this as part of our getting started guide to avoid an unnecessary line being added to the routes (which is a potential distraction for someone trying out hanami for the first time):

```
bundle exec hanami generate action home.show --skip-view --skip-routes
```

kyleplump:
@timriley would this apply to slices as well?

PR #227, opened 2024-08-26
Add --skip-route flag for action and slice generators

kyleplump:
closes [#196](https://github.com/hanami/cli/issues/196)

i tried to mirror the existing implementation of the other flags, even if we ended up with goofy functions like `Generators::App::Action#generate_route?`

lmk if you'd like to see any changes, thanks!


timriley:
The changes here look great, @kyleplump! ❤️  Could you please include one test case for the action + slice generators for when `skip_route` is passed? After that I think we can get this in!

kyleplump:
thanks @timriley! i've included the tests for your review.  lmk if this works

timriley:
Thanks @kyleplump!

PR #232, opened 2024-09-18
Update DB commands to support multiple gateways per slice

timriley:
Update the CLI for the DB gateway support introduced in https://github.com/hanami/hanami/pull/1452:

Add `--gateway` option to the following CLI commands:

- `db structure dump`
- `db structure load`
- `db create`
- `db drop`
- `db migrate`
- `db version`
- `generate migration`

These commands will operate on an individual gateway when the `--gateway` is given. Otherwise, they will operate on all the gateways for a given app or slice (or the app and all slices) based on the other args given, as before.

The above commands work with the following new file structures:

- Migrations for gateways in `config/db/[gateway_name]_migrate/` directories
- Structure files for gateways at `config/db/[gateway_name]_structure.sql`

The `db prepare` and `db seed` commands do _not_ support a `--gateway` option. This is because these commands are intended to operate on an app or slice as a whole. For example, there is only one `db/seeds.rb` file for a slice, and it should seed the tables across  the slice's gateways as appropriate.

### DIFF
--- a/lib/hanami/cli/commands/app/generate/migration.rb
+++ b/lib/hanami/cli/commands/app/generate/migration.rb
@@ -9,15 +9,35 @@ module Hanami
           # @api private
           class Migration < Command
             argument :name, required: true, desc: "Migration name"
+            option :gateway, desc: "Generate migration for gateway"
 
             example [
               %(create_posts),
               %(add_published_at_to_posts),
               %(create_users --slice=admin),
+              %(create_comments --slice=admin --gateway=extra),
             ]
 
             def generator_class
               Generators::App::Migration
+            end
+
+            def call(name:, slice: nil, gateway: nil)
+              if slice
+                generator.call(
+                  key: name,
+                  namespace: slice,
+                  base_path: fs.join("slices", inflector.underscore(slice)),
+                  gateway: gateway
+                )
+              else
+                generator.call(
+                  key: name,
+                  namespace: app.namespace,
+                  base_path: "app",
+                  gateway: gateway
+                )
+              end
             end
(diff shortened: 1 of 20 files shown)


lib/hanami/cli/commands/app/generate/migration.rb
```diff
@@ -9,16 +9,36 @@ module Generate
...
+                )
+              else
+                generator.call(
+                  key: name,
+                  namespace: app.namespace,
+                  base_path: "app",
+                  gateway: gateway
+                )
+              end
+            end
```
timriley, line 41: @cllns a note for you — to be able to pass this extra `gateway: gateway` arg when calling the generator, I had to copy the entire `#call` implementation from the base generate command class. It would be good if we could allow individual generate command classes to be able to provide extra args to the generator without having to do this. An opportunity for a refactor in the future?

spec/unit/hanami/cli/commands/app/db/structure/load_spec.rb
```diff
@@ -88,23 +88,97 @@ def db_structure_dump
       db_structure_dump
     end
 
-    def table_exists?(slice, table_name)
-      slice["db.gateway"].connection
-        .fetch("PRAGMA table_info(#{table_name})")
-        .to_a.any?
-    end
```
timriley, line 95: Being able to replace these with database-specific methods with a single `Hanami.app["db.gateway"].connection.tables.include?(:posts)` invocation is so much nicer. I wish I had explored Sequel's API a little more before I wrote these methods in the first place!

Issue #241, opened 2024-10-11
Register generators only if bundled with correct gem


swilgosz:
Currently we conditionally register the DB commands, only if `Hanami.bundled?("hanami-db")` returns true

```ruby
            if Hanami.bundled?("hanami-db")
              register "db" do |db|
                db.register "create", DB::Create
                db.register "drop", DB::Drop
                db.register "migrate", DB::Migrate
                db.register "structure dump", DB::Structure::Dump
                db.register "structure load", DB::Structure::Load
                db.register "seed", DB::Seed
                db.register "prepare", DB::Prepare
                db.register "version", DB::Version
              end
            end
```

However, our generators are always loaded.

```ruby
            register "generate", aliases: ["g"] do |prefix|
              prefix.register "action", Generate::Action
              prefix.register "component", Generate::Component
              prefix.register "operation", Generate::Operation
              prefix.register "part", Generate::Part
              prefix.register "migration", Generate::Migration
              prefix.register "relation", Generate::Relation
              prefix.register "repo", Generate::Repo
              prefix.register "slice", Generate::Slice
              prefix.register "struct", Generate::Struct
              prefix.register "view", Generate::View
            end
```

TODO:

Wrap generators registration with the `bundled?` conditional, so we can register DB or VIEW related generators only when applicable.

I've came up with quick PRs for:

- hanami-cl: https://github.com/hanami/cli/pull/242
- hanami-rspec: https://github.com/hanami/rspec/pull/26

swilgosz:
## QA

No `hanami-view`

<img width="524" alt="image" src="https://github.com/user-attachments/assets/3caf52a7-bb60-4e62-8317-92107bce6a3f">

No `hanami-db`

<img width="493" alt="image" src="https://github.com/user-attachments/assets/8ae9e252-4d0f-4e14-b55a-a58843f9031e">

No `hanami-controller`

<img width="483" alt="image" src="https://github.com/user-attachments/assets/4628da13-e2e0-48ca-abdf-f2a891b26efa">


Removed all of them:

<img width="464" alt="image" src="https://github.com/user-attachments/assets/2ca4bff4-31cd-4e99-a7ef-4c2b4e100ee6">

PR #242, opened 2024-10-11
Register generators only if bundled with correct gems

swilgosz:
Resolves: #241


swilgosz:
@timriley Thank you! Just a question: Don't we need also https://github.com/hanami/rspec/pull/26 to be merged as part of it?

timriley:
@swilgosz Thanks for the reminder! I'd overlooked that one.

PR #247, opened 2024-10-25
Run db commands on test as well as development databases

### LINKED ISSUE #215: When running database commands in development env, apply them to the test database too
Currently the user user needs to run _pairs of commands_ every time they tend to their local database:

```
hanami db create
HANAMI_ENV=test db create

hanami db migrate
HANAMI_ENV=test db migrate

# etc...
```

This is unhelpful friction, and not a great first impression on any new users.

Instead, when `db` commands are run and the `HANAMI_ENV` is `development`, then we should _also_ apply those commands to the test database.

This should take effect for the following commands:

- `db create`
- `db drop`
- `db migrate`
- `db prepare`
- `db seed`
- `structure load`

It should _not_ take effect, however, for these commands, since these only really make sense to do for a single database:

- `structure dump`
- `version`

timriley:
It's a frustrating experience if users have to remember to run each of their `hanami db` commands again with `-e test` after running them on their development database.

To improve this, update a number of commands to automatically re-run on the test database after completing their work on the development database:

- `db create`
- `db drop`
- `db migrate`
- `db prepare`
- `db structure load`

The re-run on test will only take place if these commands operate with the development environment. Running the commands in other environments (e.g. production or test) will run on the database in those environments **only.**

To achieve this "re-run on test" behaviour, at the end of these `db` commands, we invoke another process to call the relevant `hanami db` command again, but with the `HANAMI_ENV=test` environment variable set. For this re-run, all given CLI flags are preserved, but `-e` and `--env` are of course stripped.

The reason we have to take this approach is because right now there's no straightforward way to re-boot the Hanami app in a different environment, or stand up the database subsystem itself in a different environment.

I'd like to for us to make one of these possible (certainly the latter feels reasonable) in the future, but to make the user experience as good as possible for 2.2, we're taking a pragmatic approach here.

Resolves #215

### DIFF
--- a/lib/hanami/cli/commands/app/db/command.rb
+++ b/lib/hanami/cli/commands/app/db/command.rb
@@ -150,10 +150,18 @@ module Hanami
             # Invokes the currently executing `hanami` CLI command again, but with any `--env` args
             # removed and the `HANAMI_ENV=test` env var set.
             #
-            # This runs only if the Hanami env is `:development`. This is important to streamline
-            # the local development experience, making sure that the test databases are kept in sync
-            # with operations run on the development databases.
+            # This is called by certain `db` commands only, and runs only if the Hanami env is
+            # `:development`. This behavior important to streamline the local development
+            # experience, making sure that the test databases are kept in sync with operations run
+            # on the development databases.
+            #
+            # Spawning an entirely new process to change the env is a compromise approach until we
+            # can have an API for reinitializing the DB subsystem in-process with a different env.
             def re_run_development_command_in_test
+              # Only invoke a new procees if we've been called as `hanami`. This avoids awkward
+              # failures when testing commands via RSpec, for which the $0 is "/full/path/to/rspec".
+              return unless $0.end_with?("hanami")
+
               return unless Hanami.env == :development
 
               cmd = $0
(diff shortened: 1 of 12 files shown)


lib/hanami/cli/commands/app/db/command.rb
```diff
@@ -140,6 +154,66 @@ def warn_on_misconfigured_database(database, slices) # rubocop:disable Metrics/A
...
+            # `:development`. This behavior important to streamline the local development
+            # experience, making sure that the test databases are kept in sync with operations run
+            # on the development databases.
+            #
+            # Spawning an entirely new process to change the env is a compromise approach until we
+            # can have an API for reinitializing the DB subsystem in-process with a different env.
+            def re_run_development_command_in_test
+              # Only invoke a new process if we've been called as `hanami`. This avoids awkward
+              # failures when testing commands via RSpec, for which the $0 is "/full/path/to/rspec".
+              return unless $0.end_with?("hanami")
```
timriley, line 171: This is a bit hacky, but as documented, it was necessary because of our current testing approach. Given that I think we'll be able to switch to a cleaner overall approach in the future, I was comfortable with a few hacks like this being employed here.

PR #261, opened 2024-10-31
add --gateway flag to generate relation

kyleplump:
closes [#259](https://github.com/hanami/cli/issues/259)

### DIFF
--- a/lib/hanami/cli/commands/app/generate/relation.rb
+++ b/lib/hanami/cli/commands/app/generate/relation.rb
@@ -9,17 +9,37 @@ module Hanami
           # @api private
           class Relation < Command
             argument :name, required: true, desc: "Relation name"
+            option :gateway, desc: "Generate relation for gateway"
 
             example [
               %(books               (MyApp::Relation::Book)),
               %(books/drafts        (MyApp::Relations::Books::Drafts)),
               %(books --slice=admin (Admin::Relations::Books)),
+              %(books --slice=admin --gateway=extra (Admin::Relations::Books)),
             ]
 
             # @since 2.2.0
             # @api private
             def generator_class
               Generators::App::Relation
+            end
+
+            def call(name:, slice: nil, gateway: nil)
+              if slice
+                generator.call(
+                  key: name,
+                  namespace: slice,
+                  base_path: fs.join("slices", inflector.underscore(slice)),
+                  gateway: gateway
+                )
+              else
+                generator.call(
+                  key: name,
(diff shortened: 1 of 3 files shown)


lib/hanami/cli/commands/app/generate/relation.rb
```diff
@@ -9,18 +9,38 @@ module Generate
...
+                  gateway: gateway
+                )
+              else
+                generator.call(
+                  key: name,
+                  namespace: app.namespace,
+                  base_path: "app",
+                  gateway: gateway
+                )
+              end
```
timriley, line 42: Is this whole new stanza necessary just so we can pass through `gateway:`?

(I think I hit this recently too. It's something we should make better, so I just wanted to double check it's for the same reason here too)d

lib/hanami/cli/generators/app/relation.rb
```diff
@@ -19,8 +19,14 @@ def initialize(fs:, inflector:, out: $stdout)
...
           # @since 2.2.0
           # @api private
-          def call(key:, namespace:, base_path:)
+          def call(key:, namespace:, base_path:, gateway:)
             schema_name = key.split(KEY_SEPARATOR).last
+            body_content = ["schema :#{schema_name}, infer: true"]
+
+            if gateway
+              gateway_name = gateway.split(KEY_SEPARATOR).last
+              body_content.unshift("gateway :#{gateway_name}")
```
timriley, line 28:
```suggestion
              body_content.unshift("gateway :#{gateway}")
```

I don't think there's any circumstance in which gateways would have complex names with delimiters or separators, so I think this simpler approach will do just fine :)

timriley: Could we also use `.prepend` instead of `.unshift`? "Unshift" as a name truly breaks my brain, and I reckon I mustn't be the only one!

timriley (approved):
This looks great, thank you @kyleplump!

I left a couple of feedback notes for you, but they're so minor I reckon I might go and apply them directly and just merge this right in. We have a release to prepare for! 💪🏼

lib/hanami/cli/commands/app/generate/relation.rb
kyleplump: @timriley yep i believe this is necessary given the current implementation (probably looks familiar since i copied your implementation lol!) - i would be happy to start thinking of alternative ways to accomplish this

PR #264, opened 2024-11-02
Update migration link to point to Hanami website

alassek:
There's an official location for Hanami docs now, it does ultimately link back to Sequel's docs but those have a bunch of extraneous information that isn't pertinent for a Hanami project.


timriley (approved):

timriley:
So much better! Thank you!

PR #268, opened 2024-11-10
Remove ConflictingOptionsError

cllns:
Cleans up #226. 

Thanks for fixing this @timriley, it was my mistake.


timriley:
@cllns Since this class is in 2.2.0 and marked as `@api public`, I'd be inclined to leave it there. It's not causing any harm, and it's the kind of thing I could see us wanting to use in the future, anyway. What do you think?

cllns:
Yea that's fine! Sounds good

Issue #181, opened 2024-06-20
Prevent file overwriting


cllns:
For our generators, we currently overwrite a path if it already exists. We should fix this so it doesn't happen and show a helpful error message letting them know why.

For some cases (e.g. when we make a parent and a child class), we may want to output (and not overwrite) the parent class but continue on and write the child class.

I'm thinking that we can create `Hanami::CLI::Files#create` which will raise an error if the path exists and use that everywhere we need to. I guess we won't need an `update` method, since we use some of the methods on dry-files to update content at a certain place.

```ruby
# Within Hanami::CLI::Files

      def create(path, *content)
        if exist?(path)
          raise FileAlreadyExistsError.new(path) # this is new
        else
          write(path, *content)
        end
      end
```

timriley:
This is now half-resolved by #274. We no longer overwrite the file and will exit the CLI command as soon as an existing file is found.

What we need to do next, to truly fix this issue, is some kind of nicer output for the user to understand what has happened.

PR #274, opened 2024-11-14
Bugfix: Prevent generators from overwriting files

maxemitchell:
This adds a `create` method that raises an error if the file already exists.

This allows the `write` method to still be used by specs and when we want to override (e.g. for editing the routes file).

Addresses https://github.com/hanami/cli/issues/181


spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -88,7 +151,7 @@ class #{inflector.camelize(action)} < #{inflector.camelize(app)}::View
         EXPECTED
 
         expect(fs.read("app/templates/#{controller}/#{action}.html.erb")).to eq(template_file)
-        expect(output).to include("Created app/views/#{controller}/#{action}.rb")
+        expect(output).to include("Created app/templates/#{controller}/#{action}.html.erb")
```
maxemitchell, line 154: this was looking for the wrong file

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -305,47 +375,45 @@ class #{inflector.camelize(action)} < #{inflector.camelize(app)}::View
...
```
maxemitchell, line 378: same thing here with wrong file

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -305,47 +375,45 @@ class #{inflector.camelize(action)} < #{inflector.camelize(app)}::View
...
       end
     end
 
-    it "allows to specify nested action name" do
-      context = Hanami::CLI::Generators::App::ActionContext.new(inflector, app, nil, %w[api users], "thing")
-      allow(context).to receive(:bundled_views?) { true }
+    context "with nested action name" do
+      let(:context) { Hanami::CLI::Generators::App::ActionContext.new(inflector, app, nil, %w[api users], "thing") }
```
maxemitchell, line 383: added a context for this

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -1088,209 +1142,160 @@ class #{inflector.camelize(action)} < #{inflector.camelize(slice)}::View
...
+
+          template_file = <<~EXPECTED
+            <h1>#{inflector.camelize(slice)}::Views::#{inflector.camelize(controller)}::#{inflector.camelize(action)}</h1>
+          EXPECTED
+          expect(fs.read("slices/#{slice}/templates/#{controller}/#{action}.html.erb")).to eq(template_file)
+          expect(output).to include("Created slices/#{slice}/templates/#{controller}/#{action}.html.erb")
         end
       end
 
-      context "deeply nested action" do
```
maxemitchell, line 1282: this `deeply nested action` context was duplicate

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
katafrakt: nice find

timriley (approved):
Thank you for taking care of this, @maxemitchell! Especially continuing the work on it even after RubyConf finished — truly heroic! 🤩

The changes here all look good to me 👍🏼 

I'd be happy to merge this as is, because this PR alone is already a valuable improvement on what we're doing now.

Before I do that, I'd like to take this moment to get your thoughts on how this PR fits into the bigger picture.

---

With this PR in its current state, if a user ran a generator that encountered an existing file, the `hanami` CLI would crash with the new `FileAlreadyExistsError` and print out a stack trace. On the plus side, this prevents unwanted file overwriting, but the  rest of the user experience here doesn't feel so good. The error and stacktrace will be overwhelming, and it may be hard for the user to identify the cause for the CLI command crashing.

I think a better experience here might be fore our `generate` command classes to _catch_ this `FileAlreadyExistsError` and print a nicer error message before then cleanly exiting (with a non-zero exit code). This might be something we can put into a single base command class for all the generators, so that we don't have to implement it in every single command class.

What do you think about this? Any thoughts on how else this might be made nicer? And now the big question: is this something you'd also be interested in working on? 😇 I'd be very happy for it to be in a follow-up PR so that we can merge this one and stay focused on smaller, incremental changes. Let me know!

stephannv:
@timriley If you want, I can make a following PR adding a prompt confirmation:

https://github.com/user-attachments/assets/9a521df0-5821-4378-bb10-f5bfc164d342

timriley:
@stephannv that would be amazing! How about I merge this PR today, and then you can build on top of main?

stephannv:
> @stephannv that would be amazing! How about I merge this PR today, and then you can build on top of main?

Sure @timriley !

timriley:
I resolved a merge conflict to `lib/hanami/cli/generators/app/component.rb` after the changes in https://github.com/hanami/cli/pull/277.

Everything looks good now. I'll merge.

Thanks again @maxemitchell for your contribution! ❤️ Please reach out if you're interested in doing something like this again 😄

timriley:
@stephannv This is merged, feel free to base any future work off the `main` branch :)

stephannv:
@timriley while I was coding the prompt confirmation I figured out that asking the users if they want to overwrite the file may not be the best experience, this can lead to many successive questions (eg. generating an action with action, test, view, template files), and in the future we may want to add `hanami g xyz -y` to say yes to all questions, or add a option to skip (don't create the file and don't halt the generator). 

So now I'm not sure if this feature (overwrite prompt) is worth it since Hanami generators are generating simple "blank" files and not generating complex setup. 

But if we still want to overwrite existing files, maybe a `--force` option can be simple and useful, eg.
```
$ hanami g action books.index
Cannot overwrite existing file: `app/actions/books/index.rb`. Pass `--force` argument to force file overwriting.

$ hanami g action books.index --force
...
Updated app/actions/books/index.rb
...
```

PR #279, opened 2024-12-08
Exit code 1 when file already exists

### LINKED ISSUE #181: Prevent file overwriting

stephannv:
Closes #181 

@timriley I'm not sure if this message is good enough, if you have any suggestions please let me know, I made easy to change error message without having to change dozens of tests 😅.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,21 @@
 name: ci
 
-on:
+"on":
   push:
+    paths:
+      - ".github/workflows/ci.yml"
+      - "lib/**"
+      - "*.gemspec"
+      - "spec/**"
+      - "Rakefile"
+      - "Gemfile"
+      - ".rubocop.yml"
+  pull_request:
+    branches:
+      - main
+  schedule:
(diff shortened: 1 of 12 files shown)


stephannv:
Closing this because it has too many conflicts. When we want to merge this I can reopen and rebase it.

timriley:
@stephannv I'm sorry I let this drop, this is entirely my bad. 

If you can reopen it I'll attend to it asap this time.

Issue #230, opened 2024-09-02
Add --template option to generators producing view templates


Atelier-Mirai:
I like slim templates and would like to use them in HANAMI.

## --template option
The FEATURES.md file contains the following.

```FEATURES.md
### v0.8.0 - 2016-07-22
- CLI: `hanami new` can be used with `--template` argument to generate a new project with (`erb`/`haml`/`slim`) templates
```

I executed the following command.

```zsh
$ hanami new bookshelf --template=slim
ERROR: "hanami new" was called with arguments "bookshelf --template=slim"
```
Is the `--template` option already obsolete?

## generate slim template

The following command was executed to generate the action.

```zsh
$ hanami generate action home.show
```

The `app/templates/home/show.html.erb` is created.
How do I configure it to generate a show.html.slim instead of a show.html.erb?

Versions of Hanami are as follows.
```
$hanami --version
v2.2.0.beta1
```

Thank you very much for your time, and we appreciate your response.

Atelier-Mirai:
Thank you for your guidance.

I can now use slim templates by writing `gem “slim”` in the `Gemfile`.

As a newbie, I do not have the ability to issue a PR and have manually changed from `erb` to `slim`. (I just finished the tutorial on BookShelf).

I would like to donate $1 as a thank you, is it possible, I am looking forward to the completion of 2.2.0.

katafrakt:
@timriley I can take it, since there are no takers. Would you want it for 2.2 release[^1]?

Also one question for clarification:

The original report mentions `hanami new bookshelf --template=slim`, which I assume sets the default format, so it does not need to be specified every time with the generator. Should that be supported as well?

[^1]: No promises, but I can try.

Atelier-Mirai:
I am very happy to reply from you.
And thank you for working on the 2.2 implementation.

It was my understanding that each time I use slim, I need to write the template options.

Example 1

```
hanami new bookshelf --template=slim
hanami new bookshelf2 --template=slim
hanami new bookshelf3 --template=slim
```

Slim is specified as the default format. Therefore, the template option will not be necessary next time. This is very convenient and appreciated.

Example 2

```
# Set default format to slim
hanami new bookshelf --template=slim
# You don't need to specify the template option next time.
hanami new bookshelf2
hanami new bookshelf3
```

Example 2 is more convenient, but if it is difficult to implement, just Example 1 is fine.

I really appreciate your willingness to work on the implementation.
Thank you very much.

timriley:
[...]

PR #280, opened 2024-12-11
Allow specifying engine for generated templates

katafrakt:
This addresses part of #230 by allowing to specify `--template`  CLI option, which will create an ERb, HAML or Slim base template. For example:

```
hanami generate view users.sign_in.new --template haml
```


lib/hanami/cli/commands/app/generate/action.rb
```diff
@@ -51,6 +54,8 @@ class Action < App::Command
               default: DEFAULT_SKIP_ROUTE,
               desc: "Skip route generation"
             option :slice, required: false, desc: "Slice name"
+            option :template, required: false, type: :string, default: DEFAULT_TEMPLATE,
```
katafrakt, line 57: I used `template` because this was mentioned in the original issue and apparently it was named like this in Hanami 1 times. But maybe `template-engine` would be more descriptive? Although a bit long.

lib/hanami/cli/commands/app/generate/action.rb
```diff
@@ -106,7 +112,8 @@ def call(
                 raise InvalidActionNameError.new(name)
               end
 
-              generator.call(app.namespace, controller, action, url, http, format, skip_view, skip_route, slice, context: context)
+              generator.call(app.namespace, controller, action, url, http, format, skip_view, skip_route, template, slice,
```
katafrakt, line 115: It feels weird with so many positional arguments. Should I convert this to keyword?

lib/hanami/cli/commands/app/generate/view.rb
```diff
@@ -18,10 +18,10 @@ class View < App::Command
             DEFAULT_FORMAT = "html"
             private_constant :DEFAULT_FORMAT
 
-            # TODO: make engine configurable
-
             argument :name, required: true, desc: "View name"
             option :slice, required: false, desc: "Slice name"
+            option :template, required: false, desc: "Template engine to use (officially supported: erb, haml, slim)",
```
katafrakt, line 23: What should happen if an engine outside the list is called? Currently you can do `--template md` and it will create a file with `.md`  extension and with default HTML (from ERb engine). Which is not entirely wrong, but maybe the list should be restricted and produce error if user specifies unsupported engine?

This also brought the question if this should not be extensible in some way, so gems could hook into it and provide own base templates for their engines (for example `hanami-phlex`).

timriley: > maybe the list should be restricted and produce error if user specifies unsupported engine?

I think we do this for now. We can always loosen the restriction later, once we see people wanting to do more with this option.

FWIW, I see this `--template-engine` being pretty specific to the Tilt-based renderer for Hanami View. If/when a hanami-phlex gem arises, I don't think it would even use this option for that, we would have a `--render-engine` option or similar (with e.g. `tilt` or `phlex` arguments).

Issue #275, opened 2024-11-14
``Database#exists?`` is a false boolean


katafrakt:
Method checking for database existence is modeled as one returning a boolean. However, the reality is that at least three states might bi returned: exists, does not exists, we don't know.

As shown in #228 this might be problematic. Let's looks at PostgreSQL-related code:

```ruby
def exists?
  result = system_call.call("psql -t -A -c '\\list #{escaped_name}'", env: cli_env_vars)
  result.successful? && result.out.include?("#{name}|") # start_with?
end
```

If the result is not successful, for example due to permissions issue, we return `false` here. But the truth is that we **don't know** if it exists or not. The check failed, but the return value suggests that we checked and the database does not exist.

I see 3 possible ways to resolve this:

1. Keep the function as boolean, raise exception if the check failed – this is probably easiest to implement, but as with exceptions, it needs to be caught somewhere and generally breaks the execution flow
2. Rename to `check_existence`, return a Result monad. Either: `Success(:exists)`, `Success(:not_exists)`, `Failure(:check_failed)` – this would require adding `dry-monads` as a dependency of this gem, which is probably not such a big deal as Hanami in general requires `dry-monads`
3. Simplification of the above with returning an enum: `:exists`, `:does_not_exist`, `:check_failed`

kyleplump:
for what its worth: i like the exception idea

i think in this case a break of the execution flow is warranted. if there was some issue like a permissions issue, i feel as though the expectation is that execution of the action or command would stop, and that error is communicated.  so `exists?` either returns a boolean, or throws some new `Hanami::CLI::Error`

granted, the same thing could be implemented / combined with a monad response (`exec_drop_command` throws instead of `exists?`)

either way excited to see where you take it 😄

katafrakt:
Yes, I think raising exception is justified here. So if there are no other opinions, I think I'd make a PR with that change.

PR #281, opened 2024-12-11
Fail `hanami db drop` when cannot check database existence

katafrakt:
Currently if the existence check for a database fails (for example due to whole database server being down or user not existing), Hanami CLI will anyway return information that the database was dropped.

This changes it to failing with an explicit error message in such case. Examples:

```
$ hanami db drop
Could not check if the database exists. Error message:
psql: error: connection to server at "localhost" (::1), port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
connection to server at "localhost" (127.0.0.1), port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
```

```
$ hanami db drop
Could not check if the database exists. Error message:
psql: error: connection to server at "localhost" (::1), port 5432 failed: FATAL:  database "katafrakt" does not exist
```

Addresses #275

### DIFF
--- a/lib/hanami/cli/commands/app/db/utils/mysql.rb
+++ b/lib/hanami/cli/commands/app/db/utils/mysql.rb
@@ -29,8 +29,9 @@ module Hanami
               # @since 2.2.0
               def exists?
                 result = exec_cli("mysql", %(-e "SHOW DATABASES LIKE '#{name}'" --batch))
+                raise Hanami::CLI::DatabaseExistenceCheckError.new(result.err) unless result.successful?
 
-                result.successful? && result.out != ""
+                result.out != ""
               end
 
               # @api private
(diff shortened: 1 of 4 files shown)


kyleplump:
thanks for the nudge @KamilMilewski!

@katafrakt - this looks nice to me. should we implement the same behavior with the other two connection types? (`mysql`, `sqlite`)

katafrakt:
@kyleplump for SQLite the check is just `File.exist?`, so we cannot distinguish between database not existing and not being able to check it. But it seems it would make sense for MySQL. I can implement it, but first I would have to figure out installing MySQL to check it. Haven't really done it in a decade or so :sweat_smile:

kyleplump (approved):
thanks @katafrakt!

PR #284, opened 2025-01-04
Convert View generator to use RubyClassFile

cllns:
In the course of this, I realized the `RubyFileWriter` (with `#call`) is the wrong abstraction. It would create the file, but we couldn't find out any info about that class. In the case of view generation, the template shows the name of the view class, so it needs to be able to see that. In general, being able to get info from a newly generated class file seems useful in other contexts as well (imagine creating an action + operation, the action needs to know what the operation is called).

We already had an internal `ClassFile` helper inside `RubyFileWriter`, so I extracted that to its own `RubyClassFile` class, which we use in this PR. I added a `#fully_qualified_name` method too.

If we're happy with this change to the abstraction (`RubyFileWriter` -> `RubyClassFile`), I can backport the change to the other ones we've already switched. It's just a minor change and I think it's a better API (to call `#create` instead of `#call`). Since this class has always been private API (and it's an implementation detail for us, not a part of the core framework), I think it's fine to remove it during a patch release, but I'm also fine leaving it around if we're worried about backwards compatibility


timriley (approved):
This is a very useful evolution of our file generation code, @cllns! Nice work, and thank you for continuing to work in this space! I think we're heading in a good direction.

I'm very happy to see this go in. We've been intentional in keeping these facilities marked as `@api private`, so I'm not concerned about maintaining their interfaces across versions. Keeping them private is what's allowing us to evolve like this, so I think we should make the most of it and move forward assertively.

PR #293, opened 2025-03-02
Clarify customizability of config/assets.js

robyurkowski:
Previously, there was a line that was just `await assets.run()` that was not commented, and a commented block that was suggested for customization. As a result, it was possible that the initial line would be left when enabling the block, which causes a double-build and can also cause a lot of confusion once some customization is added. 

Given that the default customization function doesn't do anything, the computational cost of just executing it seems acceptable to simplify the setup and reduce the chance of frustration.

(I know this because I ran into this issue while adding Tailwind and PostCSS, and wound up being confused why builds were failing to find Tailwind. Real PEBKAC, but I hope that this would help save someone else the grief. I also expanded the comments on the file to try and tip people off that Hanami adds some configuration so that they don't also just accidentally clobber it like I did.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Hanami Command Line Interface
 
+## Unreleased
+
+### Changed
+
+- Expand on comments `config/assets.js` and enable customization function by default (@robyurkowski in #293)
+
 ## v2.2.1 - 2024-11-12
 
 ### Changed
--- a/lib/hanami/cli/generators/gem/app/assets.js
+++ b/lib/hanami/cli/generators/gem/app/assets.js
@@ -1,19 +1,17 @@
 import * as assets from "hanami-assets";
 
-/*
-  Front-end builds are powered by esbuild (https://esbuild.github.io)
-  and can be customized below. Read more at: 
-  https://guides.hanamirb.org/assets/customization/
-*/
+// Assets are managed by esbuild (https://esbuild.github.io), and can be
+// customized below.
+//
+// Learn more at https://guides.hanamirb.org/assets/customization/.
 
 await assets.run({
   esbuildOptionsFn: (args, esbuildOptions) => {
-    /* 
-      You can modify or add to Hanami's preset esbuild options here.
-      Use `args.watch` as a condition for different options for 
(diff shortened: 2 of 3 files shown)


kyleplump:
thank you for this @robyurkowski 🙂 ! I agree I think this is a nice guard rail to have with minimal cost. 

there are a few failing tests, do you mind fixing those up? then we can get this merged. Thanks!

timriley:
+1 happy to see this change go in! Let's use standard `//`-style comments rather than the special JS multi-line comments, though. These files will sit by default in a predominantly Ruby-flavoured codebase, so I think that using a consistent mode of commenting is a positive.

@kyleplump I'm happy to see you manage this the rest of the way from here :)

robyurkowski:
No probs. I'll try to get to it in the next few days, but feel free to poke me if I haven't.

timriley:
@robyurkowski Made some tweaks and merged this one in. Thanks for doing this, I think it's a real improvement!

robyurkowski:
Hey, thanks for cleaning this up and merging, Tim, and apologies I didn't get to it sooner!

One of the intents with this PR was to additionally hint that Hanami adds some configuration to esbuild such that you can't clobber `esBuildOptions` without consequence. In particular, the line `You can modify or add to Hanami's preset esbuild options here` served this function, but even that feels like it's not as clear as it could be. 

Is there a way to preserve that line, make clear that `esbuildOptions` has some modifications, or even inline those modifications so that devs aren't accidentally blowing things up? 

For example, one of the things that I first did was add `plugins: [ postCSSPlugin({ ... }) ]`, but that [overwrote](https://github.com/hanami/assets-js/blob/0e404c684780590dadd4b6768b1d7fdc9b23345d/dist/index.js#L11-L12) [the existing plugin config](https://github.com/hanami/assets-js/blob/0e404c684780590dadd4b6768b1d7fdc9b23345d/dist/esbuild.js#L60) and it seems like that would be very easy to do without an indication that `plugins` is already defined. Debugging it was a bit tricky, as I recall, and I wound up reading through source to figure out what was going on.

Any suggestions? If that feels like something that's worth an ounce of prevention, I'm happy to help where I can, though spring is a bit of a crazy season on the farm, so I might be a bit slow. :)

PR #295, opened 2025-04-12
Add simple git init at the end of new hanami app command

krzykamil:
Adding simple git repo initialization, with similar code to npm install (as for handling error display). Running it after the main install command

### DIFF
--- a/spec/unit/hanami/cli/commands/gem/new_spec.rb
+++ b/spec/unit/hanami/cli/commands/gem/new_spec.rb
@@ -1350,4 +1350,28 @@ RSpec.describe Hanami::CLI::Commands::Gem::New do
       )
     end
   end
+
+  it "initializes a git repository" do
+    expect(bundler).to receive(:install!)
+      .and_return(true)
+
+    expect(bundler).to receive(:exec)
+      .with("hanami install")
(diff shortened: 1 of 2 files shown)


spec/unit/hanami/cli/commands/gem/new_spec.rb
```diff
@@ -1350,4 +1350,28 @@ class App < Hanami::App
...
+    expect(bundler).to receive(:exec)
+      .with("hanami install")
+      .and_return(successful_system_call_result)
+
+    expect(bundler).to receive(:exec)
+      .with("check")
+      .at_least(1)
+      .and_return(successful_system_call_result)
+
+    expect(system_call).to receive(:call).with("npm", ["install"])
```
timriley, line 1367: The fact that we have to have all these non-relevant setup lines just so we can test for `git init` feels like a bit of a small to me, and suggests we should probably improve the test setup for the `new` command at least. Doing that in this PR would make it much larger than it needs to be, so I'm OK for this to test to be as it is for the sake of keeping this PR to a smaller, focused change :)

timriley (approved):
Thanks for this, @krzykamil! Looks good to me :)

PR #298, opened 2025-04-30
Detect slice when generating

krzykamil:
This PR introduces automatic slice detection when user runs a generator from inside a slice. Based on simple `pwd` output.

The detection happens in the parent Command class in the `call`, that is used by `super` by children commands.

The specs for this have been added to just the parent class and its tested as a method spec for all commands that end up using it indirectly. This is sort of a "compromise" between having a separate spec for each command (lots of maintenance involved, little benefit) and having just the spec for this method for the parent class. So it basically checks if children commands all can use this method.

Maybe another spec checking if it is used from the `call` method could be useful


cllns:
This will implement #185, is that correct?

krzykamil:
> This will implement #185, is that correct?

Hey, I was about to ask for your opinion :)

Yes that is the issue I am trying to solve. I'm wondering if it is a good direction (a simple module inserted into generators), and if adding a test for each command is right, after testing the module itself?

Needs a rebase cause I started it on a train few weeks back and there were some changes to the commands I modified since then, will deal with this soon

krzykamil:
I've changed the approach to instead leverage generators inheriting from 
`Hanami::CLI::Commands::App::Generate` and calling `super` in their `call`s
I've added one example spec to `action` spec to see if this seems to be working. But not sure if the test is very good, cause I had to stub a lot to simulate user calling the generator from the slice subdirectory.

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
```diff
@@ -1345,6 +1345,37 @@ def handle(request, response)
...
+          allow(mock_app).to receive(:inflector).and_return(inflector)
+
+          allow(subject).to receive(:app).and_return(mock_app)
+
+          slice_path = File.join(Dir.pwd, "slices", slice)
+          allow(Dir).to receive(:pwd).and_return(slice_path)
+
+          subject.call(name: action_name)
+
+          expect(fs.exist?("slices/#{slice}/actions/#{controller}/#{action}.rb")).to be(true)
```
krzykamil, line 1373: Those checks are happening a lot in the codebase. Wondering if a verbose, custom rspec matcher would do some good on those?
```ruby
expect(an_action_to_exist(action, slice: slice))
```
and such
or even, with some custom chain methods
```ruby
expect(an_action_to_exist(action).in_slice(slice))
```

I know not everyone is a fan of custom rspec matches, but I quite like them, coming from a place where I think RSpec is a big, heavy tool, that by default is just a base spec framework to me modified in according to project needs. Leveraging those custom matchers etc. is to me one of the main reasons to use RSpec and writing specs instead of tests using some smaller test framework.

lib/hanami/cli/commands/app/generate/command.rb
```diff
@@ -38,9 +38,23 @@ def generator_class
...
+              current_dir = Pathname.new(Dir.pwd)
+              app_root = app.root
+              slices_dir = app_root.join("slices")
+              return nil unless current_dir.to_s.start_with?(slices_dir.to_s)
+
+              relative_path = current_dir.relative_path_from(slices_dir)
+              slice_name = relative_path.to_s.split("/").first
+              return slice_name if app.slices[slice_name.to_sym]
+
+              nil
```
cllns, line 51:
```suggestion
              slice_name if app.slices.keys.include?(slice_name.to_sym)
```
We can use implicit return instead of explicit (also returning nil when condition is false). 

Additionally, `app.slices[...]` raises a Hanami::SliceLoadError when it's not found, so we need to check on the keys. Can you add tests for this path? Likely by keeping the `app.slices[...]` and ensuring the error is raised, then modifying the code so it doesn't raise anymore.

lib/hanami/cli/commands/app/generate/command.rb
```diff
@@ -38,9 +38,23 @@ def generator_class
               # fs:, inflector:, out:
             end
 
+            def detect_slice_from_pwd
+              current_dir = Pathname.new(Dir.pwd)
+              app_root = app.root
+              slices_dir = app_root.join("slices")
+              return nil unless current_dir.to_s.start_with?(slices_dir.to_s)
```
cllns, line 45:
```suggestion
              return unless current_dir.to_s.start_with?(slices_dir.to_s)
```

lib/hanami/cli/commands/app/generate/command.rb
```diff
@@ -38,9 +38,23 @@ def generator_class
               # fs:, inflector:, out:
             end
 
+            def detect_slice_from_pwd
+              current_dir = Pathname.new(Dir.pwd)
+              app_root = app.root
```
cllns, line 43:
```suggestion
```
Only used once so not worth storing the variable

lib/hanami/cli/commands/app/generate/command.rb
```diff
@@ -38,9 +38,23 @@ def generator_class
               # fs:, inflector:, out:
             end
 
+            def detect_slice_from_pwd
+              current_dir = Pathname.new(Dir.pwd)
```
cllns, line 42:
```suggestion
              current_dir = Pathname.pwd
```

lib/hanami/cli/commands/app/generate/command.rb
```diff
@@ -38,9 +38,23 @@ def generator_class
...
+              slice_name = relative_path.to_s.split("/").first
+              return slice_name if app.slices[slice_name.to_sym]
+
+              nil
+            end
+
             # @since 2.2.0
             # @api private
             def call(name:, slice: nil, **opts)
+              slice ||= detect_slice_from_pwd
```
cllns, line 57: I wonder if we should do this in the constructor instead?

Also, do we want to allow generating files in a slice `foo` when we're inside a slice `bar`? I'm leaning towards preventing `--slice` and `--app` when inside a slice or app/ but what do other think?

spec/unit/hanami/cli/commands/app/generate/action_spec.rb
cllns: Hmm that's an interesting idea! I say leave it for now, but let's keep it in mind as we build this out.

cllns (changes requested):
Thanks for doing this, and rebasing! I have some notes about the PR but the overall approach seems solid. 

I guess we'll want to add specs for each of the commands? Or maybe have a generic command spec, so we don't have quite so much duplication?

lib/hanami/cli/commands/app/generate/command.rb
krzykamil: cool, did not know about this

lib/hanami/cli/commands/app/generate/command.rb
krzykamil: Hard to say, I'd say that if someone is in a slice, and wants to generate something for some other place (slice or main app) and they have to go out of the slice they are in, they might be annoyed (not a big hindrance but still). The `pwd` option for me is a shortcut for user, but explicit option overwriting the default would be the preferred behavior for me.

krzykamil:
> I guess we'll want to add specs for each of the commands? Or maybe have a generic command spec, so we don't have quite so much duplication?

Yea thats my main problem. Adding a spec for each command for this is most thorough but then if something changes with this logic and we have to modify each of them... quite a chore. Maybe I'll add a separate "GenerateCommandBasedOnPwd" (not final name)  and in there I'll write a spec that iterates over all commands and runs the same spec for all of them that checks if this thing works correctly. Think I like that the most, will post it in upcoming days

lib/hanami/cli/commands/app/generate/command.rb
krzykamil: Done

krzykamil:
@cllns I've changed the approach making the change only 2 files now (although the spec I am not 100% happy with). Care to take a look when you get the time?

lib/hanami/cli/commands/app/generate/command.rb
timriley: Great discussion, folks. Here's my opinion:

### `#call` or `#initialize` for pwd check?

I think this pwd-based slice detection logic makes more sense to stay in `#call`.

`#initialize` is for dependencies. `#call` is for “doing the work.” To me, checking for the current working directory is part of the latter.

Even though our CLI invocations tend to be one-shot, I could imagine a world where e.g. some kind of REPL-like experience is built on top of CLI commands. In that kind of scenario, command objects may be longer-lived, and I'd want “current state” checks to be done when a command is invoked (which, in this hypothetical scenario, may be done many times) rather than initialised. (which would be done once).

I know this is a really subtle distinction, and it can be changed easily enough in the future either way, but in this case I'm arguing for the case of “change nothing”, which feels easy enough to follow :)

### role of explicit `--slice` or `--app` flags when executed within a slice

I think the `--slice` and `--app` flags should work (and be respected) regardless of where you invoke the CLI. The falls under “principle of least surprise” IMHO. The flags will always exist for the case of executing from the root. They'll be documented. We should make sure people can depend on them, which means respecting them at all times.

If we made the flags work in some cases and not others, the CLI experience would feel more “slippery” than dependable, and we don't want that.

If a user is supplying these flags, it's because they've taken the time to type them out. They know what they want. We should respect it.

lib/hanami/cli/commands/app/generate/command.rb
```diff
@@ -38,9 +38,22 @@ def generator_class
...
+            def detect_slice_from_pwd
+              current_dir = Pathname.pwd
+              slices_dir = app.root.join("slices")
+              return unless current_dir.to_s.start_with?(slices_dir.to_s)
+
+              relative_path = current_dir.relative_path_from(slices_dir)
+              slice_name = relative_path.to_s.split("/").first
+              return unless app.slices.keys.include?(slice_name.to_sym)
+
+              slice_name if app.slices[slice_name.to_sym]
```
timriley, line 50: Worth noting here that we technically allow nested slices, which I'd like to make sure we support it in anything that interacts with slices.

The approach in this method detects slices in the top-level `slices/` directory only. Nested slices OTOH will be in paths like `slices/admin/slices/editorial/` (see https://github.com/hanami/hanami/blob/c1efc98f2b813274920da30c73e6d71ba75705b8/spec/integration/slices/slice_loading_spec.rb for some examples).

So I wonder if a more robust approach here is to do something like this?

```
slices_by_root = app.slices.with_nested.each.to_h { |slice| [slice.root, slice] }
slices_by_root[current_dir]
```

Slices are already loaded as part of accessing the `app`, so it's no extra work to enumerate their roots at this stage. Plus, this code feels more straightforward and the implementation should be more reliable even if our handling of slice roots changes in the future. What do you reckon?

spec/unit/hanami/cli/commands/app/command_spec.rb
```diff
@@ -17,4 +19,41 @@ def call(**)
...
+
+  context "#detect_slice_from_pwd" do
+    let(:out) { StringIO.new }
+    let(:fs) { Hanami::CLI::Files.new(memory: true, out: out) }
+    let(:inflector) { Dry::Inflector.new }
+    let(:slice_registrar) { instance_double(Hanami::SliceRegistrar) }
+
+    context "for all commands" do
+      %w[View Struct Repo Operation Component Action Part Migration Relation].each do |cmd|
+        it "doesn't crash when slice does not exist" do
```
timriley, line 29: Thanks for all the thought you've been putting into how to test this, @krzykamil.

The approach we have here feels a little too low-level. It's testing an internal private method rather than the behaviour that it is meant to enable.

How about this as an approach?

Make a new `spec/unit/hanami/cli/commands/app/generate/slice_detection_spec`. This spec file will be dedicated to the slice detection feature only. It can contain tests that look like the other `generate` command tests, but it can focus on the slice detection feature only, and to do this, it only needs to exercise one generator command (view or action or whatever you choose); we can just trust that the slice detection works for all the others rather than writing heaps of redundant extra tests.

The tests that go into this file don't need to be as detailed as the other command-specific tests in the neighbouring files. It just needs to check that the right slice has been picked based on the cwd.

So I think it'd look something like this:

```ruby
it "detects the slice based on the current working directory" do
  within_application_directory do
    fs.mkdir("slices/main")

    Dir.chdir("slices/main") do
      subject.call(name: "users.index")
    end

    expect(fs.exist?("slices/main/views/users/index.rb")).to be true
  end
end
```

Something like that. Just enough to know the slice was detected.

Then in the rest of this new test file we can test the few other scenarios we need to cover, like when we're in a slice directory but explicitly provide a different slice as an arg, when we're inside a deeper app directory that is not a slice directory, etc. Probably not too many in total.

How does that sound? Does it make sense? Happy to clarify anything :)

timriley (changes requested):
This is looking good, thanks @krzykamil! I've left:

- Some [answers for you](https://github.com/hanami/cli/pull/298#discussion_r2232677382) on the open questions
- A suggestion for more reliably matching slice roots
- A suggestion for the tests

We're nearly there! I don't think any of these should require too much work :)

Thank you again for all your work on Hanami! And I'm really sorry for my delay in replying here.

spec/unit/hanami/cli/commands/app/command_spec.rb
krzykamil: Thanks, it seems i need to use `fs.pwd` (which I didn't knew was an option) and figure out some other things. I started building the new spec and it is a better way, just need a bit of time to figure out how it should all look like

timriley:
Have pushed a little spike of an alternative approach! https://github.com/hanami/cli/pull/314

krzykamil:
> Thanks @krzykamil! I made a few adjustments (see last commit) and now it's merged!
> 
> I also added a big long comment explaining why we have to make two checks for the slice root:
> 
> ```ruby
>             # Returns the root for the given slice name.
>             #
>             # This currently works with top-level slices only, and it simply appends the slice's
>             # name onto the "slices/" dir, returning e.g. "slices/main" when given "main".
>             #
>             # TODO: Make this work with nested slices when given slash-delimited slice names like
>             # "parent/child", which should look for "slices/parent/slices/child".
>             #
>             # This method makes two checks for the slice root (building off both `app.root` as well
>             # as `fs`). This is entirely to account for how we test commands, with most tests using
>             # an in-memory `fs` adapter, any files created via which will be invisible to the `app`,
>             # which doesn't know about the `fs`.
>             #
>             # FIXME: It would be better to find a way for this to make one check only. An ideal
>             # approach would be to use the slice_name to find actual slice registered within
>             # `app.slices`. To do this, we'd probably need to stop testing with an in-memory `fs`
>             # here.
>             def detect_slice_root(slice_name)
>               slice_root_in_fs = fs.join("slices", inflector.underscore(slice_name))
>               return slice_root_in_fs if fs.exist?(slice_root_in_fs)
> 
>               app.root.join("slices", inflector.underscore(slice_name))
>             end
> ```
> 
> It makes me think that we probably just need to stop using the in-memory `fs` adapter for CLI command tests. The Hanami app itself doesn't know about or use this `fs` object, which (as you discovered!) makes it very difficult to write commands that integrate with facilities returned from the `Hanami.app`.

Thanks Tim, just a question:

This comment
            # TODO: Make this work with nested slices when given slash-delimited slice names like
            # "parent/child", which should look for "slices/parent/slices/child".
            
I had this spec 
```ruby
            context "with deeply nested slices existing" do
      it "determines the nested slice when inside of it" do
        Dir.chdir("slices/admin/slices/search") do
          subject.call(name: "panel")
        end

        expect(File.exist?("slices/admin/slices/search/views/panel.rb")).to be true
      end
```

Which I guess is passing. Doesn't it meet the case from the comment? Or am I missing something?

timriley:
@krzykamil Ah yes, good question! So when I said "make it work with nested slices", what I meant was _not_ the detection of the nested slice based on the pwd (which already works, as you point out), but _rather_ being able to specify a nested slice via the `--slice` flag.

In other words, make this work:

```
hanami generate view panel --slice parent/child
# generates slices/parent/slices/child/views/panel.rb
```

Does that make sense?

This is not something we have supported so far, but with this code change being made, it felt like a sensible place to be able to note this as a TODO for the future :)

PR #300, opened 2025-06-20
Rollback command

krzykamil:
1. Focused on doing things 1 database per rollback, no "assuming" behavior user would want with multiple databases
2. Enforce specific usage of the command, unless the setup is very basic or very obvious (single app, single database etc.)
3. Use err.puts to communicate what user needs to do (which I based on it being used in lib/hanami/cli/commands/app/db/command.rb)
4. Utilize command_exit more (looked more into the migrate command and based it on that)
5. Hope the messages to the user are clear enough, lemme know WDYT
6. Provided a little comment about calculating target_index

If this is accepted I will provide relevant documentation changes

### DIFF
--- a/lib/hanami/cli/commands/app/db/rollback.rb
+++ b/lib/hanami/cli/commands/app/db/rollback.rb
@@ -1,6 +1,4 @@
 # frozen_string_literal: true
-
-require "pry"
 require_relative "../../app/command"
 require_relative "structure/dump"
 
@@ -18,6 +16,9 @@ module Hanami
             option :gateway, required: false, desc: "Use database for gateway"
 
             def call(steps: nil, app: false, slice: nil, gateway: nil, target: nil, dump: true, command_exit: method(:exit), **)
+              # We allow either a number of steps or a target migration number to be provided
+              # If steps is provided and target is not, we use steps as the target migration number, but we also have to
+              # make sure steps is a number, hence some additional logic around checking and converting to number
               target = steps if steps && !target && !code_is_number?(steps)
               steps_count = steps && code_is_number?(steps) ? Integer(steps) : 1
 
@@ -78,7 +79,7 @@ module Hanami
               return unless slice
 
               databases = build_databases(slice)
-              
+
               if gateway
                 database = databases[gateway.to_sym]
                 unless database
--- a/spec/unit/hanami/cli/commands/app/db/rollback_spec.rb
+++ b/spec/unit/hanami/cli/commands/app/db/rollback_spec.rb
@@ -1,8 +1,47 @@
 # frozen_string_literal: true
 
-require "pry"
-
 RSpec.describe Hanami::CLI::Commands::App::DB::Rollback, :app_integration do
+  shared_examples "single database rollback" do |db_type|
+    it "rolls back the most recent migration" do
+      columns = -> { Hanami.app["db.gateway"].connection.schema(:posts).map(&:first) }
+      expect(columns.()).to eq [:id, :title, :body, :published]
+
+      command.call(app: true, gateway: "default")
+
+      expect(columns.()).to eq [:id, :title, :body]
+      expect(output).to include "rolled back"
+      expect(dump_command).to have_received(:call).with(hash_including(app: true)).once
+    end
+  end
+
+  shared_examples "multiple gateways error handling" do |db_type|
+    it "fails with clear error message when multiple gateways exist without specification" do
+      exit_mock = instance_double("Method", call: nil)
+      allow(exit_mock).to receive(:call).with(1).and_raise(SystemExit)
+      
+      expect { command.call(command_exit: exit_mock) }.to raise_error(SystemExit)
+      expect(output).to include "Multiple gateways found in app. Please specify --gateway option."
+    end
+  end
+
+  shared_examples "invalid argument handling" do
+    it "fails when gateway is specified without app or slice" do
+      exit_mock = instance_double("Method", call: nil)
+      allow(exit_mock).to receive(:call).with(1).and_raise(SystemExit)
+      
+      expect { command.call(gateway: "default", command_exit: exit_mock) }.to raise_error(SystemExit)
+      expect(output).to include "When specifying --gateway, an --app or --slice must also be given"
+    end
+
+    it "fails when gateway does not exist" do
+      exit_mock = instance_double("Method", call: nil)
+      allow(exit_mock).to receive(:call).with(1).and_raise(SystemExit)
+      
+      expect { command.call(app: true, gateway: "nonexistent", command_exit: exit_mock) }.to raise_error(SystemExit)
+      expect(output).to include %(No gateway "nonexistent" found in app)
+    end
+  end
+
   subject(:command) {
     described_class.new(
       system_call: system_call,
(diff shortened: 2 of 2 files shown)


lib/hanami/cli/commands/app/db/rollback.rb
```diff
@@ -0,0 +1,175 @@
...
+
+            private
+
+            def rollback_across_all_databases(steps:, dump:, gateway:, command_exit:)
+              all_databases = databases(app: false, slice: nil, gateway: gateway)
+
+              # Collect all applied migrations across all databases with their source database
+              all_migrations = []
+              all_databases.each do |database|
+                applied_migrations = database.applied_migrations
```
kyleplump, line 66: might be able to memoize this -  `all_migrations_by_database` hash or something. would save a call to `applied_migrations` on line 91

lib/hanami/cli/commands/app/db/rollback.rb
```diff
@@ -0,0 +1,175 @@
...
+
+              migrations_to_rollback.each do |migration_info|
+                database = migration_info[:database]
+                migration_name = migration_info[:name]
+
+                # Find the previous migration in this database
+                all_db_migrations = database.applied_migrations
+                current_index = all_db_migrations.index { |m| m.include?(migration_info[:migration]) }
+
+                next if current_index.nil? # This shouldn't happen, but I am not like 100% sure?
```
kyleplump, line 94: i agree i don't think this should happen, _but_ i like keeping here, better safe than sorry! :smile: 

could this even be some sort of error message instead of silently skipping? would that be useful, or exposing internal implementation too much?

lib/hanami/cli/commands/app/db/rollback.rb
```diff
@@ -0,0 +1,175 @@
...
+                # If steps exceed available migrations, rollback all migrations
+                # by using the first (oldest) migration
+                migration =
+                  if index < -applied_migrations.size
+                    return initial_state(applied_migrations)
+                  else
+                    applied_migrations[index]
+                  end
+              else
+                migration =
```
kyleplump, line 145: `migration` could be `nil` here. a user is supplying the timestamp, could mistype and the `detect` could not find anything :frowning:

lib/hanami/cli/commands/app/db/rollback.rb
```diff
@@ -0,0 +1,175 @@
...
+              target = steps if steps && !target
+
+              if !app && slice.nil? && (steps.nil? || (steps && code_is_number?(steps)))
+                steps_count = steps.nil? ? 1 : Integer(steps)
+                rollback_across_all_databases(steps: steps_count, dump: dump, gateway: gateway, command_exit: command_exit)
+              else
+                databases(app: app, slice: slice, gateway: gateway).each do |database|
+                  migration_code, migration_name = find_migration(target, database)
+
+                  if migration_name.nil?
```
kyleplump, line 30: could `migration_name` ever end up being `nil?`?

lib/hanami/cli/commands/app/db/rollback.rb
```diff
@@ -0,0 +1,175 @@
...
+                    output = if target
+                               "==> migration file for target #{target} was not found"
```
kyleplump, line 32: we might need to be a little more clever with the output here. if `steps` was given and not `target`, `target` could end up being like '2' or something (since its being initialized to `steps` on line 21), which could be confusing

kyleplump (changes requested):
awesome work @krzykamil!! :muscle:  i left a few comments 

more broadly: do you think its worth adding a flag for the number of steps? e.g.: `rollback -s 3` or `rollback --steps=2` or something similar.  i like the simplicity of just passing a number, but i feel the explicitness could both prevent mistypes, help simplify the implementation (since its currently handling the ambiguity), and reduce overall 'magic'

maybe we could keep just passing a single number (like we currently have), as a shortcut. and then the `steps` flag would be 'in-addition'?  i'm curious what you and the rest of the team think about this!

looking forward to seeing where this goes! this is a great feature to have

lib/hanami/cli/commands/app/db/rollback.rb
krzykamil: unsure either, so I left it as `next` but if someone sees a strong incentive to have it as error I will change it :+1:

lib/hanami/cli/commands/app/db/rollback.rb
```diff
@@ -0,0 +1,175 @@
...
+
+            def find_migration(code, database)
+              applied_migrations = database.applied_migrations
+
+              return if applied_migrations.empty?
+
+              # Rollback to initial state if we have only one migration and
+              # no target is specified. In this case the rollback target
+              # will be the current migration timestamp minus 1
+              return initial_state(applied_migrations) if applied_migrations.one? && code.nil?
```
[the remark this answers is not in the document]
krzykamil: If no arguments are passed to steps or target it could happen, so when the default behavior is present.

lib/hanami/cli/commands/app/db/rollback.rb
krzykamil: Correct, I will add a test case for that too, to handle user input mistake better. Thanks!

krzykamil: true, need to cover that in a spec too I guess

lib/hanami/cli/commands/app/db/rollback.rb
krzykamil: when there are no migrations to rollback for example, or when user inout makes a imstake

lib/hanami/cli/commands/app/db/rollback.rb
krzykamil: In this branch user provided steps not a timestamp. If more steps are provided than thee are available we rollback to initial state

krzykamil: added a spec for that and made me realize I missed a spec where I pass the timestamp

kyleplump:
great, thanks @krzykamil!

everything else looks good to me, curious what everyone else thinks 🙂

krzykamil:
> great, thanks @krzykamil!
> 
> everything else looks good to me, curious what everyone else thinks 🙂

Thanks, thanks

I am also wondering how to do specs in a smart way. Not sure if just copying every spec I have into context for postgres and mysql is the best way, or should I DRY it up (personally not a big fan of RSpec shared_example as it makes debugging nasty sometimes).

If none has better ideas I would just copy the specs for each of the database adapters

alassek:
@krzykamil 

> But I kinda did not understood what the app flag is for except if we would want to command to infer the slice from the place the command is run

The default behavior, if `--app` or `--slice` are absent, is to execute against _all_ gateways. The `--app` flag limits the migrator to just the app-level db.

See https://github.com/hanami/cli/blob/fa4aeed23868a8f9dad4e68e59999c09198f391f/lib/hanami/cli/commands/app/db/command.rb#L62-L69

krzykamil:
> What do you reckon?

You make valid points ofc, I did work out some scenarios without thinking how useful they are (was thinking too much about if I could, rather than if I should :d .
Rolling back more than 1 database at a time seems very edge-casey/uncommon and counter-intuitive. Keeping in mind consequences of 1 rollback is enough, doubt someone will bother doing that in bigger scale

I will refocus on this PR now with those things in mind:
1. `--app` option means "run this only for main app context" (this will be the new default behavior if there are no multiple databases)
  a. If there are multiple databases (across app or slices or gateways) and neither `app` or `slice` is provided, user will be prompted for one or the other
  b. If there are multiple gateways in a given context (app has multiple gateways or slice), user will be prompted to specify which one is to be chosen
2. Only allow the command to work on one database at a time, for both user and implementation simplicity

krzykamil:
This has been updated, I am still working on making specs for other databases but the new approach should be clearly visible now

PR #305, opened 2025-07-08
Disable foreman's `.env` processing

### LINKED ISSUE #304: `hanami dev`  handles .env differently
I tend to run development modes with an isolated database within a docker `compose.yml`.  the compose isn't important, but, for general non-clashyness, I don't bind to a specific port.

Docker gives you a mechanism for finding the mapping for the current port, for example:
```
$ docker compose port db 5432
0.0.0.0:32771
```

This can be used inside a `dotenv` loaded environment file to set the database_url, and hanami will read this successfully:
```
$ cat .env 
# This is checked into source control, so put sensitive values into `.env.local`
DATABASE_URL=postgres://user:pass@$(docker compose port db 5432)/demo

$ hanami server
09:42:55 - INFO - Using Guardfile at /home/flip/src/demo/Guardfile.
09:42:56 - INFO - Puma starting on port 2300 in development environment.
09:42:56 - INFO - Guard is now watching at '/home/flip/src/demo'
Puma starting in single mode...
* Puma version: 6.6.0 ("Return to Forever")
* Ruby version: ruby 3.4.4 (2025-05-14 revision a38531fd3f) +PRISM [x86_64-linux]
*  Min threads: 5
*  Max threads: 5
*  Environment: development
*          PID: 339648
* Listening on http://0.0.0.0:2300
* Starting control server on http://127.0.0.1:9293
* Starting control server on http://[::1]:9293
Use Ctrl-C to stop
```

However, when using `hanami dev` this results in a stack trace:
```
09:44:10 web.1    | ! Unable to load application: URI::InvalidURIError: bad URI (is not URI?): "postgres://user:pass@$(docker compose port db 5432)/demo"
09:44:10 web.1    | /opt/rubies/3.4.4/lib/ruby/3.4.0/uri/rfc3986_parser.rb:130:in 'URI::RFC3986_Parser#split': bad URI (is not URI?): "postgres://user:pass@$(docker compose port db 5432)/demo" (URI::InvalidURIError)
09:44:10 web.1    | 	from /opt/rubies/3.4.4/lib/ruby/3.4.0/uri/rfc3986_parser.rb:135:in 'URI::RFC3986_Parser#parse'
```

This seems to be because foreman loads `.env` on it's own, and is not acutally fully compatible with all of what dotenv offers.
https://github.com/ddollar/foreman/issues/702

There _does_ appear to be an option to avoid this:  `--env=/dev/null`  ... I think that `hanami-cli` should be updated to use this option so that the `.env` file is processed via the same mechanism regardless of how it's launched.

cflipse:
fixes #304

### DIFF
--- a/lib/hanami/cli/generators/gem/app/dev
+++ b/lib/hanami/cli/generators/gem/app/dev
@@ -5,4 +5,4 @@ if ! gem list foreman -i --silent; then
   gem install foreman
 fi
 
-exec foreman start -f Procfile.dev "$@"
+exec foreman start -f Procfile.dev --env=/dev/null "$@"
--- a/spec/unit/hanami/cli/commands/gem/new_spec.rb
+++ b/spec/unit/hanami/cli/commands/gem/new_spec.rb
@@ -238,7 +238,7 @@ RSpec.describe Hanami::CLI::Commands::Gem::New do
           gem install foreman
         fi
 
-        exec foreman start -f Procfile.dev "$@"
+        exec foreman start -f Procfile.dev --env=/dev/null "$@"
       EXPECTED
       expect(fs.read("bin/dev")).to eq(bin_dev)
       expect(fs.executable?("bin/dev")).to be(true)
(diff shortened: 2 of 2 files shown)


timriley:
Thanks @cflipse!

PR #307, opened 2025-07-09
Inject exit command to Create in tests

### LINKED ISSUE #306: Problem with running tests
When I try to run the tests for this repo locally, sometimes my test run looks like this:

```
Randomized with seed 7379
................................................................................................*..................................mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
WARNING: option --ssl-verify-server-cert is disabled, because of an insecure passwordless login.


Pending: (Failures listed here are expected and do not affect your suite's status)

  1) Hanami::CLI::Commands::App::Generate::Action generate for app, without hanami view bundled allows to specify MIME Type for template
     # Temporarily skipped with xit
     # ./spec/unit/hanami/cli/commands/app/generate/action_spec.rb:252


Finished in 8.57 seconds (files took 0.85412 seconds to load)
132 examples, 0 failures, 1 pending
```

Other times like this:

```
Randomized with seed 6645
...................mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
WARNING: option --ssl-verify-server-cert is disabled, because of an insecure passwordless login.


Finished in 4.46 seconds (files took 0.84088 seconds to load)
20 examples, 0 failures
````

Either way, the whole suite is pretty much never completed. It looks like something (probably around mysql tests) kills the test process abruptly, but in a way that the RSpec runner thinks that all the test are run. I'm not sure how to start debugging it. Running with `doc` format does not reveal any further hints.

Example:

```

Randomized with seed 15437

Hanami::CLI::Server
  #call
    delegates config as config
    delegates debug as debug
    delegates port as Port
    delegates warn as warn
    delegates host as Host

Hanami::CLI::Commands::App::DB::Structure::Load
  mysql
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
WARNING: option --ssl-verify-server-cert is disabled, because of an insecure passwordless login.

Finished in 0.32933 seconds (files took 1.25 seconds to load)
6 examples, 0 failures
```

Another example (truncated)

```
Randomized with seed 49833

bin/hanami
  console
    defaulting to pry when it's present
      starts pry console
    irb
      starts irb console (PENDING: Temporarily skipped with xit)
[...]
Hanami::CLI::Commands::App::DB::Structure::Dump
  postgres
    prints errors for any dumps that fail and exits with non-zero status
    dumps the structure for the app db when given --app
    dumps the structure for each db, including schema_migrations
    dumps the structure for a slice db when given --slice
  sqlite
    prints errors for any dumps that fail and exits with non-zero status
    dumps the structure for each db, including schema_migrations
    dumps the structure for the app db when given --app
    dumps the structure for a slice db when given --slice
    slice with gateways
      dumps the structure for for a slice's gateway when given --slice and --gateway
      dumps the structure for all the slice's gateways when given --slice
    app with gateways
      dumps the structure for all the app's gateways when given --app
      dumps the structure for for a slice's gateway when given --app and --gateway
  mysql
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
WARNING: option --ssl-verify-server-cert is disabled, because of an insecure passwordless login.

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) bin/hanami console irb starts irb console
     # Temporarily skipped with xit
     # ./spec/integration/run_spec.rb:59


Finished in 9.42 seconds (files took 0.55687 seconds to load)
47 examples, 0 failures, 1 pending
```

**And a somewhat related question:**

Could setup for running the tests be documented somewhere in the repository, perhaps the README? I find myself reverse-engineering Github Actions setup every time, just to figure out how to run against different databases and perhaps I'm doing something wrong in that regard.

The command I used to run above is:

```
MYSQL_BASE_URL=mysql2://katafrakt@127.0.0.1/hanami_cli_test POSTGRES_BASE_URL=postgres://postgres:password@localhost:5432/hanami_cli_test bundle exec rake spec
```

(tried replacing with `bundle exec rspec` but the outcome is the same)

### PR DESCRIPTION
**Pull request by katafrakt**
There was an issue with test runs ending abruptly with no message when a database configuration was incorrect. In such cases the Create command ran an `exit(1)`, which exited not only the command, but the whole test run.

The solution was already prepared, but not used: it was already possible to inject an alternative to just calling the actual `Kernel.exit`. This injects a test double instead and asserts it is not called.

Fixes #306

### DIFF
--- a/lib/hanami/cli/commands/app/db/drop.rb
+++ b/lib/hanami/cli/commands/app/db/drop.rb
@@ -11,7 +11,7 @@ module Hanami
 
             option :gateway, required: false, desc: "Use database for gateway"
 
-            def call(app: false, slice: nil, gateway: nil, **)
+            def call(app: false, slice: nil, gateway: nil, command_exit: method(:exit), **)
               exit_codes = []
 
               databases(app: app, slice: slice, gateway: gateway).each do |database|
@@ -27,7 +27,7 @@ module Hanami
               end
 
               exit_codes.each do |code|
-                break exit code if code > 0
+                break command_exit.(code) if code > 0
               end
 
               re_run_development_command_in_test

### CODE REVIEW

**Comment by cllns** [comment]
Excellent! Confirmed this works. Getting 317 examples evaluated every time locally now. 

Can we get rid of [the existing code](https://github.com/search?q=repo%3Ahanami%2Fcli%20%23%20Prevent%20the%20command%20from%20exiting%20the%20spec%20run%20in%20the%20case%20of%20unexpected%20system%20call%20failures&type=code) that tried to solve this issue?
```
    # Prevent the command from exiting the spec run in the case of unexpected system call failures
    allow(command).to receive(:exit)
```
Or could we somehow make that work, since it's simpler?

I see that same line is used in the prepare_spec too, which isn't in this PR. Do we need to extend this fix to that file as well?

katafrakt:
This existing code is actually in use, at least for `Drop` command ([see here](https://github.com/hanami/cli/blob/712a1ad19913993435989d59e894b648c59f541d/lib/hanami/cli/commands/app/db/drop.rb#L30)) - when I remove it, tests start exiting again. Perhaps `Drop` should follow the same pattern as other commands and call injected exit instead of just calling `exit`?

katafrakt:
I modified the `Drop` command to look and behave like the other commands in the PR. However, I still cannot remove the "old `allow", as tests start to fail.

I agree that just calling `exit` from the command class feel a bit aggressive and perhaps some other mechanism should be used, however maybe it's meant to be like that in Dry::CLI.

As for `break`, if the injected command does not raise `SystemExit`, it probably should break anyway, so I would keep it.

Issue #302, opened 2025-06-21
Hanami new --api mode


cllns:
It'd be great if we could support JSON API's from `hanami new` without the user needing to do any additional configuration.

It'd include --skip-assets, and I think it should also --skip-view (which doesn't exist yet: #301). Beyond that, we should configure actions to accept and return JSON only, and remove all Capybara/feature specs files.

I think the CLI fla should be `--api [json]`, to provide an extension point for people who want to generate other kinds of API's but provide `json` as the default since that'll be the majority of uses.

PR #308, opened 2025-07-10
Add app level --skip-view flag

### LINKED ISSUE #301: Hanami new --skip-view
Not everyone will want to use our view layer, since they may want to render JSON responses directly from their actions, or use a different view/templating layer (like Phlex) directly.

We already have --skip-view for actions, but this would be at the project level.

kyleplump:
closes #301 

This PR adds an optional `--skip-view` flag to the `hanami new` command. including this flag will not bundle the `hanami-view` gem, and will not generate any of the initial view files or directories (e.g. the things in `app/views` and `app/templates`)

however, this _does_ still potentially allow assets files to be created.  the idea being if you wanted to skip views and assets you'd use `hanami new` in api mode (#302)

### DIFF
--- a/lib/hanami/cli/commands/gem/new.rb
+++ b/lib/hanami/cli/commands/gem/new.rb
@@ -32,6 +32,11 @@ module Hanami
 
           # @since 2.2.0
           # @api private
+          SKIP_VIEW_DEFAULT = false
+          private_constant :SKIP_VIEW_DEFAULT
+
+          # @since 2.2.0
+          # @api private
           DATABASE_SQLITE = "sqlite"
 
           # @since 2.2.0
@@ -81,6 +86,12 @@ module Hanami
 
           # @since 2.2.0
           # @api private
+          option :skip_view, type: :flag, required: false,
+                             default: SKIP_VIEW_DEFAULT,
+                             desc: "Skip including hanami-view"
+
+          # @since 2.2.0
+          # @api private
           option :database, type: :string, required: false,
                             default: DATABASE_SQLITE,
                             desc: "Database adapter (supported: sqlite, mysql, postgres)"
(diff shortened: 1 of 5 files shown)


kyleplump:
hi @cllns :wave:

i have two questions related to this, hopefully you could provide some insight:
- for skipping views, should we still allow the default `public/404.html` and `public/500.html` files?  
- from what i can tell, there currently is not logic to raise an error when trying to run the `hanami generate view ..` command, _without_ the gem being bundled.  is that something you'd like to have? are we planning on supporting the view generator with alternative view layers?

thanks as always! :cherry_blossom:

timriley:
@kyleplump hello! Allow me to furnish you with some answers 🧐 

> for skipping views, should we still allow the default public/404.html and public/500.html files?

Let's remove them. I think that if someone asks to `--skip-views`, they'll be surprised if these files remain in their newly generated Hanami app.

Technically, these files aren't rendered by Hanami View or any related part of the view layer. They're instead shown by [a middleware](https://github.com/hanami/hanami/blob/fa46f546a58d0fb76b065f1d7d543223a5cbb3ed/lib/hanami/middleware/render_errors.rb) when the app's `config.render_errors` is set to `true`. The error handling invoked by that middleware [gracefully handles the absence of the files](https://github.com/hanami/hanami/blob/fa46f546a58d0fb76b065f1d7d543223a5cbb3ed/lib/hanami/middleware/public_errors_app.rb#L64-L72), so there'll be no issues if they're not around.

> from what i can tell, there currently is not logic to raise an error when trying to run the hanami generate view .. command, without the gem being bundled. is that something you'd like to have?

These `generate view` commands should disappear once the hanami-view gem is no longer bundled :) [See here](https://github.com/hanami/cli/blob/fa4aeed23868a8f9dad4e68e59999c09198f391f/lib/hanami/cli/commands/app.rb#L55-L58). If you're seeing something different, please let us know!

timriley:
The code changes in this PR all look great to me, so once we skip generating the public/ error files, this will be good to merge!

kyleplump:
thanks for the answers @timriley! i committed the change to skip generating those html error files as well.

> These generate view commands should disappear once the hanami-view gem is no longer bundled :) [See here](https://github.com/hanami/cli/blob/fa4aeed23868a8f9dad4e68e59999c09198f391f/lib/hanami/cli/commands/app.rb#L55-L58). If you're seeing something different, please let us know!

i didn't know this, thanks for clarifying!

PR #312, opened 2025-07-31
Use app inflector for app commands

### LINKED ISSUE #299: Hanami generate does not respect inflections
Lets say you have:

```ruby
# frozen_string_literal: true

require 'hanami'

module MyApp
  class App < Hanami::App
    config.inflections do |inflections|
      inflections.acronym 'IDP'
    end
  end
end
```

And you do any of the following:

```sh
hanami generate slice idp
hanami generate action 'sessions.new' --slice idp
```

You would expect the generated action to be:

```ruby
# frozen_string_literal: true

module IDP
  module Actions
    module Sessions
      class New < IDP::Action
        def handle(request, response)
        end
      end
    end
  end
end
```

However both the slice files, and the action are generated with the name `Idp`.  This is very frustrating as it means you're always having to correct the generated files.

timriley:
These commands were previously using the generic inflector we were making available via the base command. This means that running `generate` commands inside an app would end up creating e.g. classes whose names did not use the app's custom inflection rules.

With this change, remove the inflector as an injectable dependency of commands. This wasn’t providing any benefit for testing, and it wasn’t necessary for the commands to function. It also ensures the app commands use the right inflector at all times (the app inflector), since it is no longer substitutable.

Fixes #299

### DIFF
--- a/spec/integration/hanami/cli/commands/app/generate/part_spec.rb
+++ b/spec/integration/hanami/cli/commands/app/generate/part_spec.rb
@@ -4,16 +4,11 @@ require "hanami"
 require "tempfile"
 
 RSpec.describe "Hanami generate part integration", :app do
-  let(:fs) { Hanami::CLI::Files.new(memory: false) }
-  let(:inflector) { Dry::Inflector.new }
+  let(:fs) { Hanami::CLI::Files.new(memory: false, out: out) }
   let(:out) { StringIO.new }
 
   subject(:command) do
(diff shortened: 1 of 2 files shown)


timriley:
@cllns — are you OK with the removal of the injectable inflector? I don't consider this a breaking change because our CLI app command classes isn't really widely used by third parties, and we haven't documented how to use them at a third party level either. In addition, a third party is likely inheriting and just treating all the default `#initialize` args as `**` and passing them on, rather than naming them.

PR #313, opened 2025-08-01
Fix ArgumentError for Part generator

cllns:
Clean up for #292. Right now `hanami generate part` (from hanami/cli main) is broken.
 
`skip_tests:` wasn't being accepted, so it was raising an ArgumentError (hanami-rspec looks at that option)

This also adds a regression test for that, plus allowing arbitrary kwargs args


timriley (approved):
Thanks @cllns :)

PR #318, opened 2025-09-15
add missing automatic rake tasks loading

AlexanderZagaynov:
This pull request introduces a small but useful improvement to the Hanami application's Rake configuration. The change ensures that any custom rake task placed in the `lib/tasks` directory will be automatically loaded.

### DIFF
--- a/spec/unit/hanami/cli/commands/gem/new_spec.rb
+++ b/spec/unit/hanami/cli/commands/gem/new_spec.rb
@@ -215,6 +215,9 @@ RSpec.describe Hanami::CLI::Commands::Gem::New do
         # frozen_string_literal: true
 
         require "hanami/rake_tasks"
+
+        # Add your custom rake tasks to the lib/tasks directory
+        Rake.add_rakelib "lib/tasks"
       EXPECTED
       expect(fs.read("Rakefile")).to eq(rakefile)
       expect(output).to include("Created Rakefile")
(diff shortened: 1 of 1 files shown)


AlexanderZagaynov:
@alassek So far `rake` is still the part of hanami dependencies, and used quite intensively: [[1]](https://github.com/hanami/cli/blob/d6f965106a2071a920a1e2a54461d243d1aafe43/hanami-cli.gemspec#L38), [[2]](https://github.com/hanami/cli/blob/d6f965106a2071a920a1e2a54461d243d1aafe43/lib/hanami/cli/rake_tasks.rb), it's even a part of the generated app itself: [[3]](https://github.com/hanami/cli/blob/d6f965106a2071a920a1e2a54461d243d1aafe43/lib/hanami/cli/generators/gem/app/rakefile.erb), so I think it's definitely a bit premature to think about `rake` abandonment. My motivation was that I wasted some of my time trying to understand why my task can't be found and run, and wanted to save such time for a person behind me.

timriley:
Yep, I think this is a good thing to add. Thank you!

I agree that it would be nice to make Hanami CLI extensions the easy first choice for users wanting to add their own CLI tasks, but the fact is that we have a Rakefile generated by default, so if a user is comfortable using Rake, then we shouldn't get in their way. (They're already taking a chance on a bunch of new things by trying Hanami in the fish place, Rake feels like an area where we can help them continue to use familiar tools).

So I'm happy to see this go in, it feels like a very manageable load for us to carry :)

I'll merge this when I get a chance. 

Thank you for the PR, and thanks @alassek for weighing in. I do think it's healthy that we question these things :)

PR #324, opened 2025-09-29
Consume App configuration for console

alassek:
Closes hanami/hanami#1538
Depends on console config introduced in hanami/hanami#1540


timriley (approved):
Beautiful!

As a last step, would you mind adding something to point this out in our 2.3 guides? (The folder for this version is already ready to go in our guides repo).

Thanks Adam!

PR #336, opened 2025-10-06
Remove restrict and unrestrict from pg dumps

rickenharp:
This change removes the noise introduced in recent postgres versions by adding a new unique key every time the schema gets dumped.

### DIFF
--- a/lib/hanami/cli/commands/app/db/utils/postgres.rb
+++ b/lib/hanami/cli/commands/app/db/utils/postgres.rb
@@ -44,7 +44,15 @@ module Hanami
                 system_call.call(
                   "pg_dump --schema-only --no-privileges --no-owner --file #{structure_file} #{escaped_name}",
                   env: cli_env_vars
-                )
+                ).tap do
+                  next unless it.successful?
+                  dump_lines = File.readlines(structure_file)
+                  File.open(structure_file, "w") do |f|
+                    dump_lines.each do |line|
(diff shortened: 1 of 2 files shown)


lib/hanami/cli/commands/app/db/utils/postgres.rb
```diff
@@ -44,7 +44,15 @@ def exec_dump_command
                 system_call.call(
                   "pg_dump --schema-only --no-privileges --no-owner --file #{structure_file} #{escaped_name}",
                   env: cli_env_vars
-                )
+                ).tap do
+                  next unless it.successful?
```
alassek, line 48: Be mindful of what language features you use in library code. This is unsupported by 3/4 of the allowed versions in the gemspec.

timriley:
> I believe all three database types support dumping to stdout, so we could introduce api to the `database` object to post-process the dump output and write to the filesystem.

Yep, this sounds like a good move to me.

I'd be happy if this was as a fast-follow to this PR, so at least we can keep each one focused on a single change.

timriley:
I think that remove the `it` here will get us green on all versions and ready to merge.

@rickenharp thanks for tackling this! Do you reckon you might be up for doing what @alassek suggested as a follow-up PR, processing the lines in-memory before writing to the file?

rickenharp:
I just rewrote this change on top of #337 and force-pushed, that should fix the tests.

PR #342, opened 2025-10-12
Appends routes to slice routes file if it exists

### LINKED ISSUE #341: Action generation should handle routes file under the slice directory
I'm using [this setup](https://guides.hanamirb.org/v2.2/app/slices/#slice-routing):
> If you want to separate the routes of your slice from those of the application, you can cut the routes from config/routes.rb and create a new file under the slice directory: slices/admin/config/routes.rb
> ```ruby
> # config/routes.rb
> module MyApp
>   class Routes < Hanami::Routes
>     root { "Hello from Hanami" }
> 
>     slice :admin, at: "/admin"
>   end
> end
> 
> # slices/admin/config/routes.rb
> module Admin
>   class Routes < Hanami::Routes
>     ...
>   end
> end
> ``` 

When I run `hanami g action foo.bar --slice=admin`, the new route is added to `config/routes.rb` instead of `slices/admin/config/routes.rb`, and I have to move the new route manually to the correct file.

stephannv:
Fixes #341.

Check if the slice has a `slices/xyz/config/routes.rb` file. If it exists, append the route there; otherwise, keep the default behavior by adding it to the slice block in `config/routes.rb`.

### DIFF
--- a/lib/hanami/cli/generators/app/action.rb
+++ b/lib/hanami/cli/generators/app/action.rb
@@ -83,8 +83,14 @@ module Hanami
             if namespace == Hanami.app.namespace
               fs.inject_line_at_class_bottom(routes_location, "class Routes", route)
             else
-              slice_matcher = /slice[[:space:]]*:#{namespace}/
-              fs.inject_line_at_block_bottom(routes_location, slice_matcher, route)
+              slice_routes = fs.join("slices", namespace, "config", "routes.rb")
+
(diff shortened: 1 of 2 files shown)


afomera (approved):
Thanks @stephannv!! This fix will help improve the experience of those using slices with a dedicated routes file 🙌.

Pulled this down and tested it in a project locally too and worked like a charm!

PR #346, opened 2025-10-21
Prefer Bundler.with_original_env over Bundler.with_unbundled_env.

robyurkowski:
`InteractiveSystemCall` / `SystemCall` both use `Bundler.with_unbundled_env { }`, which clobbers any bundler environment variables you might have set. For example, it's not uncommon to set `BUNDLE_PATH` in a Docker environment.

This winds up causing some super confusing errors, like when you run `hanami db migrate`, the development db is migrated, but then when it tries to do the test environment, it crashes and tells you that it can't find your gems anywhere:

image

The solution is dead simple: just switch to `Bundler.with_original_env`:

image

In this PR, I've also expanded the specs around `SystemCall` / `InteractiveSystemCall` since they were under-specced.

### DIFF
--- a/lib/hanami/cli/interactive_system_call.rb
+++ b/lib/hanami/cli/interactive_system_call.rb
@@ -13,13 +13,12 @@ module Hanami
         @out = out
         @err = err
         @exit_after = exit_after
-        super()
       end
 
       # @api private
       # @since 2.1.0
       def call(cmd, *args, env: {}, out_prefix: "")
-        ::Bundler.with_unbundled_env do
+        ::Bundler.with_original_env do
           threads = []
           exit_status = 0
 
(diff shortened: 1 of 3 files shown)


lib/hanami/cli/interactive_system_call.rb
```diff
@@ -13,13 +13,12 @@ def initialize(out: $stdout, err: $stderr, exit_after: true)
         @out = out
         @err = err
         @exit_after = exit_after
-        super()
```
robyurkowski: This class doesn't inherit from anything, so this is a bit superfluous. I thought maybe this was designed to inherit from `SystemCall` (and I think there's a pretty decent overlap, so that seems like a reasonable refactoring), but this is only conjecture.

timriley (approved):
This makes a lot of sense. Thanks for finding and fixing this issue, @robyurkowski! Appreciate all the extra test coverage too ❤️ 

We'll release this with Hanami 2.3 in a couple of weeks.

PR #349, opened 2025-10-24
Print warnings for unbooted slices in console

timriley:
Add another console “plugin” to overload `Slice.keys` and print a one-time-per-slice warning when accessing keys for an unbooted slice.

This “plugin” is slightly differently-shaped to the `SliceReaders` one we have currently, but it needs to be because its purpose is not actually to enhance the console context, but rather to patch some external code. I think this difference is fine, and in the future we can move to make a more formal definition of what it means to be a console plugin.

This is an alternative take to @kyleplump’s https://github.com/hanami/hanami/pull/1543. @kyleplump, the reason I like this approach better is that it doesn’t add these console-specific behaviours to `Hanami::Slice` that complicate things in all the times that it’s _not_ running within a console (i.e. most of the time!). It also keeps this specific behaviour closer to the thing that needs it in the first place: the console classes themselves.

How does this feel to you?

After I added the code to make this print one-time only, I think we can probably get away without having the config to disable these notices altogether. Let's at least see how we go and learn from user feedback about whether it makes sense to add the config.

Fixes https://github.com/hanami/hanami-cli/issues/325

### DIFF
--- a/lib/hanami/console/context.rb
+++ b/lib/hanami/console/context.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative "plugins/slice_readers"
+require_relative "plugins/unbooted_slice_warnings"
 
 module Hanami
   # @since 2.0.0
@@ -21,6 +22,8 @@ module Hanami
 
         define_context_methods
         include Plugins::SliceReaders.new(app)
+
+        Plugins::UnbootedSliceWarnings.activate
       end
 
       private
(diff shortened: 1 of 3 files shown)


timriley:
Thanks for checking this out, @kyleplump! I'll merge this one in now (and for anyone else who looks, I'm very happy to receive post-merge feedback!) and we can consider this little enhancement complete :) Thanks for your work in pushing it forward!

PR #350, opened 2025-10-24
Generate a View Context class when creating a new application

### LINKED ISSUE #345: Generate view context class in new apps

afomera:
This generates app/views/context.rb when generating a new Hanami application by default 

Fixes #345.

This should help introduce the context concept to new users and give them the way to add methods without needing to create the file themselves.


timriley:
Thanks @afomera! I just tweaked the comment to include a URL to the relevant guides page. Bringing this in!

PR #360, opened 2025-10-31
Always return a result from database command methods

brkn:
### Description

- Return `SystemCall::Result` objects from guard checks at database util class methods instead of primitive `true` values. This removes the need for checking the object of the returned.

- Replace `Failure` Struct with a class to remove warnings at test outputs. `Failure constant redefined`


brkn:
Thank you for the comment.

> I’m not sure I like the specific solution of faking shell command output.

To understand it better, which one you don't like: 

A) These database methods (like `exec_drop_command`) returning `SystemCall::Result` instances when we didn't do system calls on early exits

B) These database methods (like `exec_drop_command`) returning an object that respond to `#exit_code` calls when we didn't do system calls on early exits

PR #372, opened 2025-11-28
Remove explicit `preload_app!` from Puma config

joshuay03:
As of [Puma 7.0](https://github.com/puma/puma/blob/main/History.md#700--2025-09-03), preloading is enabled by default in cluster mode, so there's no need to explicitly enable it.

This is a slightly opinionated change to remove the explicit enabling of a config that's now enabled by default, so feel free to close this if not preferred.

### DIFF
--- a/lib/hanami/cli/generators/gem/app/gemfile.erb
+++ b/lib/hanami/cli/generators/gem/app/gemfile.erb
@@ -22,7 +22,7 @@ source "<%= gem_source %>"
 
 gem "dry-types", "~> 1.7"
 gem "dry-operation", ">= 1.0.1"
-gem "puma"
+gem "puma", ">= 7.1"
 gem "rake"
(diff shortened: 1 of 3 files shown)


timriley:
Hey @joshuay03, always nice to see a PR from you!! 😄

I think this change is a good one. It keeps us up to date, and it slims down our standard config. Win win!

The only thing to consider is that we should make sure the user is actually using Puma 7 or higher, otherwise we risk not preloading the app if a user bundles and already has an older Puma installed. So we should add a ">= 7" version constraint to the generated Gemfile in new apps. Do you reckon you could expand this PR to cover that?

joshuay03:
FYI Rails will be requiring the latest version (7.1), let me know if you want to do the same: https://github.com/rails/rails/pull/56254

timriley:
Yeah, let's require latest.

PR #375, opened 2026-01-07
Prepare for repo-sync

timriley:


.github/workflows/repo-sync-extensions/ci-services.yml
```diff
@@ -0,0 +1,25 @@
...
+  ports:
+    - 3307:3306
+  options: >-
+    --health-cmd "mysqladmin ping"
+    --health-interval 10s
+    --health-timeout 5s
+    --health-retries 3
+postgres:
+  # Use postgres:14 for CLI compatibility with ubuntu-latest, currently ubuntu-22.04
+  # See https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md
```
katafrakt: ubuntu-latest is actually 24.04 now and it supports PostgreSQL 16 by default. Perhaps we can bump it here (14 is really old).

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#postgresql

PR #378, opened 2026-01-21
Fix CI run on JRuby

katafrakt:
The tests success for JRuby is optional, but the CI currently fails in setup, which makes the CI red. This just fixes this one issue (plus adds one thing to env vars), but does not make the JRuby tests pass - the latter is a lot of works and probably should be taken in few smaller steps. Meanwhile, the issue with red CI is problematic.

### DIFF
--- a/.github/workflows/repo-sync-extensions/ci-env.yml
+++ b/.github/workflows/repo-sync-extensions/ci-env.yml
@@ -1,1 +1,2 @@
 POSTGRES_BASE_URL: postgres://postgres:password@localhost:5432/hanami_cli_test
+JRUBY_OPTS: "-X+O"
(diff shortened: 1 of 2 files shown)


timriley:
Thank you so much, @katafrakt! ❤️

.github/workflows/repo-sync-extensions/ci-env.yml
```diff
@@ -1 +1,2 @@
 POSTGRES_BASE_URL: postgres://postgres:password@localhost:5432/hanami_cli_test
+JRUBY_OPTS: "-X+O"
```
timriley: What's this one for? (And should we be standardising on this across all our CI?)

PR #379, opened 2026-01-21
Replace exits with throws in commands

katafrakt:
I'm seeing test runs ending prematurely again (see #307) locally. This is my attempt to address this problem more systematically, which happened to use the same approach that Tim suggested in the last comment in the other PR.

### DIFF
--- a/exe/hanami
+++ b/exe/hanami
@@ -7,9 +7,14 @@ require "hanami/cli"
 cli = Dry::CLI.new(Hanami::CLI)
 Hanami::CLI::Bundler.require(:cli)
 
-begin
-  cli.call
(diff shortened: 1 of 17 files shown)


timriley:
This is a great improvement, thank you @katafrakt! I'm very sorry for the delay in my review, but I think we can merge this now, and because it's an internal change only, release it with the next release.

I also took a quick look at the other repos where we add or enhance our CLI commands (like hanami-rspec and hanami-reloader), and it doesn't look like any changes are needed there.

timriley (approved):

katafrakt:
No worries about the delay. This is just to ensure that we are always running all specs, so we don't have accidental false positives (happened to me in the past).

PR #381, opened 2026-02-04
Use assets config path relative to app root in assets command

haileys:
A little context first because this is a weird one: I have a somewhat ..unorthodox node.js setup on my dev machines where `node`, `npm`, etc. are wrapper scripts that transparently run node stuff in a container where only the essentials are mounted in. Usually this just works, and surprisingly well too, but it does mean that workspace path node sees is not the same as the workspace path outside of the container. Most stuff I run seems to use relative paths so this hasn't really been an issue.

The `hanami assets` command invokes node with an absolute path however, and my weird setup breaks. I did a bit of digging and it looks like assets-js [already assumes cwd == app.root](https://github.com/hanami/hanami-assets-js/blob/main/src/index.ts#L17), so I think it should be safe also assume cwd == app.root here in the assets command, and invoke node with a path relative to the app root.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 ### Changed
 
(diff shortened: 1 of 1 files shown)


haileys:
Thank you!

PR #387, opened 2026-02-19
Add `--name` option to `hanami new` command

aaronmallen:
Allows users to specify a custom module namespace for the generated app while using the positional argument as the directory path.

For example, `hanami new my_bookshelf --name=bookshelf` creates the app in `my_bookshelf/` with `Bookshelf` as the module name.

### DIFF
--- a/lib/hanami/cli/commands/gem/new.rb
+++ b/lib/hanami/cli/commands/gem/new.rb
@@ -107,6 +107,10 @@ module Hanami
                             default: DATABASE_SQLITE,
                             desc: "Database adapter (supported: sqlite, mysql, postgres)"
 
+          # @api private
+          option :name, type: :string, required: false,
+                        desc: "App name to use for the module namespace"
+
           # rubocop:disable Layout/LineLength
           example [
             "bookshelf                                    # Generate a new Hanami app in `bookshelf/' directory, using `Bookshelf' namespace",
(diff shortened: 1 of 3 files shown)


lib/hanami/cli/commands/gem/new.rb
```diff
@@ -147,17 +152,19 @@ def call(
...
 
-            raise PathAlreadyExistsError.new(app) if fs.exist?(app)
+            raise PathAlreadyExistsError.new(directory) if fs.exist?(directory)
             raise ForbiddenAppNameError.new(app) if FORBIDDEN_APP_NAMES.include?(app)
 
             normalized_database ||= normalize_database(database)
 
-            fs.mkdir(app)
-            fs.chdir(app) do
+            fs.mkdir(directory)
```
katafrakt: I think maybe this should be `mkdir_p`, in case someone does `--path my/bookshelf/app`?

katafrakt (commented):
Looks good. As a general comment, I know some frameworks chose to rather have `--app-name` option or similar. This allows to avoid pitfalls of nested paths, relative vs absolute etc.

In this case it would be `hanami new my_bookshelf --app-name bookshelf`. But I think it's fine either way.

timriley:
Cool, thank you! Please add a CHANGELOG entry and then merge when you're ready :)

PR #389, opened 2026-02-21
Make generators respect config.views.default_template_engine

katafrakt:
Second part of #230 

This allows Hanami application to set a default template engine (https://github.com/hanami/hanami/pull/1564) and CLI will pick that config up - and will use it unless other template engine was specified via the CLI option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 - Add `--name` option to `hanami new`, to specify a custom module namespace while using the positional argument as the directory path. For example, `hanami new my_bookshelf --name=bookshelf` creates the app in `my_bookshelf/` with `Bookshelf` as the module name. (@aaronmallen in #387)
 - Add `--template-engine` option to `hanami generate`, to specify which template engine should be used for generated template file. Supports `erb`, `haml` and `slim`. (@katafrakt in #280)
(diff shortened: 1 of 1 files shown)


lib/hanami/cli/commands/app/generate/view.rb
```diff
@@ -16,11 +16,13 @@ module Generate
...
 
+            DEFAULT_TEMPLATE_ENGINE = "erb"
+            private_constant :DEFAULT_TEMPLATE_ENGINE
+
             argument :name, required: true, desc: "View name"
             option :template_engine, required: false,
-                                     desc: "Template engine to use",
                                      values: %w[erb haml slim],
-                                     default: "erb"
+                                     desc: "Template engine to use (default: set in application config or erb)"
```
timriley:
```suggestion
                                     desc: "Template engine to use (default: set in app config or erb)"
```

I try for us to use the shorthand "app" consistently, since that's what we name our app class and file.

lib/hanami/cli/commands/app/generate/view.rb
```diff
@@ -32,6 +34,26 @@ class View < Command
             def generator_class
               Generators::App::View
             end
+
+            # @since 2.0.0
+            # @api private
+            def call(name:, slice: nil, template_engine: nil, **opts)
+              super(
+                name: name,
+                slice: slice,
```
timriley:
```suggestion
                name:,
                slice:,
```

Let's adopt this implicit hash value syntax wherever we can :)

If you could make this equivalent change up in the generate/action command too, that would be great!

PR #402, opened 2026-04-21
Update references to hanami-action

cllns:
Renamed from hanami-controller

See: https://github.com/hanami/hanami/pull/1582

I updated all references to 'controller' in the codebase too.

I also changed the gemfile sources to reflect the new naming scene of `hanami/hanami-<gem>` from just `hanami/<gem>`. I only _had_ to update the hanami-action one, but did the others for consistency.


timriley (approved):
Thanks @cllns!

("Namespace" is a much better term than the "controller" term we were previously using. This is a good move all around!)

timriley:
I hope you don't mind, @cllns, but I'll just merge this directly, so main branch CI becomes green again :)

PR #404, opened 2026-05-01
Use threads instead of forking to compile slices assets

katafrakt:
One of the main blockers for JRuby adoption in Hanami is that Hanami CLI uses `fork` to spawn assets compilation for slices in parallel. I was trying to find a reason why it was done like that for a while, but could not find any. I also recently found https://github.com/hanami/hanami/pull/1524#issuecomment-2977063644 which asks about the same and no conclusion was made.

This PR replaces `Process.fork` with `Thread.new`. In my tests everything still works and assets are being compiled in parallel.

### DIFF
--- a/lib/hanami/cli/commands/app/assets/command.rb
+++ b/lib/hanami/cli/commands/app/assets/command.rb
@@ -11,8 +11,8 @@ module Hanami
         module Assets
           # Base class for assets commands.
           #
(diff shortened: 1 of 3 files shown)


timriley (approved):
Happy to see this go in! I'd love to get JRuby support locked in for our next Hanami release :)

PR #405, opened 2026-05-01
Include `rouge` gem in `Gemfile` for new apps

### LINKED ISSUE #394: Add rouge gem to standard new app Gemfile (unless --skip-db)
This is to enable SQL syntax highlighting in logs. See https://github.com/hanami/hanami/pull/1570.

kyleplump:
This PR includes the `rouge` gem in the default `Gemfile` for new apps.  It will not include the gem if `skip-db` flag is passed

Closes [#394](https://github.com/hanami/hanami-cli/issues/394)

### DIFF
--- a/lib/hanami/cli/generators/gem/app/gemfile.erb
+++ b/lib/hanami/cli/generators/gem/app/gemfile.erb
@@ -37,9 +37,6 @@ gem "haml"
 <%- if generate_view? && template_engine == "slim" -%>
 gem "slim"
 <%- end -%>
-<%- if generate_db? -%>
-gem "rouge"
-<%- end -%>
 
 group :development do
   <%= hanami_gem("webconsole") %>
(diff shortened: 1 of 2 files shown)


katafrakt (approved):

timriley (approved):
Thanks @kyleplump!

Couple of little adjustments:

- Can we move this down into the `:test, :development` group? We don't need syntax highlighting (nor the related dependency) in production.
- This could do with a comment above it, since I think users will not be able to infer why it's there. Something simple like "Syntax highlighting SQL logs" would be enough.

Once you're done with those, happy for you to merge :)

kyleplump:
I debated where it should go, thank you for the direction @timriley!  Both of your requested changes have been implemented

timriley:
Perfect, thank you @kyleplump!

PR #414, opened 2026-06-02
Fix blank Postgres CLI env values

### LINKED ISSUE #412: db tasks fail when connecting to Postgres with a socket in non-standard location
## Describe the bug

Due to the CLI tasks for Postgres overriding `PGHOST` and other `libpq` env vars, using a `DATABASE_URL` value like `postgres:///development` only works if the socket is stored in `/tmp`.

## To Reproduce

The hardest part of this bug report would be telling others how to configure a Postgres in this way. In my case, I'm using devenv, which sets the following `PG_*` vars:

```
PGPORT=5432
PGHOST=/tmp/devenv-2921247/postgres
PGDATA=/Users/bjeanes/Code/gms/gms-hanami/.devenv/state/postgres
```

However, when used with `DATABASE_URL=postgres:///gms`:

```sh-session
❯ bundle exec hanami db create
Could not check if the database exists. Error message:
psql: error: connection to server on socket "/tmp/.s.PGSQL.5432" failed: No such file or directory
        Is the server running locally and accepting connections on that socket?
```

## Expected behavior

Connection should be established using the correct socket path.

When connecting with a socket, `libpq` uses `PGHOST` as the directory which holds the socket and `PGPORT` to build the socket name.

Ideally, this code https://github.com/hanami/hanami-cli/blob/b36827f73a3de4cdc639b0cb80218af346bac407/lib/hanami/cli/commands/app/db/utils/postgres.rb#L89-L96 would consider socket-based connections.

I considered opening a pull request (and am still happy to do so), but noted that there are no specs covering this existing behaviour so wanted first to sanity check the approach.

It seems like that code may always have a truthy value for `database_uri.host` in this case, unlike the other fields:

```sh-session
❯ irb -ruri

⢀⡴⠊⢉⡟⢿  IRB v1.18.0 - Ruby 3.4.9
⣎⣀⣴⡋⡟⣻  "debug" to start integration with the "debug" gem
⣟⣼⣱⣽⣟⣾  ~/Code/gms/gms-hanami

irb(main):001> URI.parse("postgres:///some-database").host
=> ""
irb(main):002> URI.parse("postgres:///some-database").port
=> nil
irb(main):003> URI.parse("postgres:///some-database").user
=> nil
irb(main):004> URI.parse("postgres:///some-database")..password
=> nil
```

So the easiest patch may be to handle the empty string as falsey for these purposes.

Alternatively, perhaps it makes more sense never to clobber `PG*` vars when the calling environment already sets them, using a pattern such as: `vars["PGHOST"] = ENV.fetch("PGHOST") { ... }`.

Lastly, the usages of these Postgres tools do support a full URL (since about 9.2, IIRC):

```
psql -d "postgres://user:pass@localhost:5433/mydb"
pg_dump -d "postgres://user:pass@localhost:5433/mydb" -f dump.sql
createdb -d "postgres://user:pass@localhost:5433/" newdb
dropdb -d "postgres://user:pass@localhost:5433/" mydb
```

So, the full URI could just be passed there instead of parsing them into `PG*` vars. However, `createdb` and `dropdb` use these as the connection URI so in those cases the database name at least would also need to be passed as a separate argument.

## My environment

- Ruby version: 3.4.9
- OS: macOS Tahoe 26.5
- devenv 2.1.2
- Postgres 18

StantonMatt:
Closes #412.

`postgres:///db_name` socket-style URLs can leave URI connection fields blank. The db CLI currently turns those blank fields into `PG*` env values, which means `PGHOST=""` can override a caller-provided socket directory and make the Postgres tools fall back to the default socket path.

This keeps the existing behavior for non-empty host/port/user/password values, but skips blank URI fields so libpq can use the surrounding environment. I also added a small Postgres utility spec covering both non-empty values and blank values.

I checked the focused Postgres utility spec, RuboCop on the touched files, Ruby syntax for the touched files, and whitespace before opening this.

### DIFF
--- a/lib/hanami/cli/commands/app/db/utils/postgres.rb
+++ b/lib/hanami/cli/commands/app/db/utils/postgres.rb
@@ -87,19 +87,10 @@ module Hanami
               end
 
               def cli_env_vars
-                @cli_env_vars ||= {}.tap do |vars|
-                  add_cli_env_var(vars, "PGHOST", database_uri.host)
-                  add_cli_env_var(vars, "PGPORT", database_uri.port)
-                  add_cli_env_var(vars, "PGUSER", database_uri.user)
-                  add_cli_env_var(vars, "PGPASSWORD", database_uri.password)
+                @cli_env_vars ||= %i[host port user password].each_with_object({}) do |field, vars|
+                  value = database_uri.public_send(field).to_s
+                  vars["PG#{field}".upcase] = value unless value.empty?
                 end
-              end
-
-              def add_cli_env_var(vars, name, value)
-                value = value.to_s
-                return if value.empty?
-
-                vars[name] = value
               end
             end
           end
--- a/lib/hanami/cli/commands/app/db/utils/postgres.rb
+++ b/lib/hanami/cli/commands/app/db/utils/postgres.rb
@@ -86,13 +86,20 @@ module Hanami
                 Shellwords.escape(name)
               end
 
-              def cli_env_vars # rubocop:disable Metrics/AbcSize
+              def cli_env_vars
                 @cli_env_vars ||= {}.tap do |vars|
-                  vars["PGHOST"] = database_uri.host.to_s if database_uri.host
-                  vars["PGPORT"] = database_uri.port.to_s if database_uri.port
-                  vars["PGUSER"] = database_uri.user.to_s if database_uri.user
-                  vars["PGPASSWORD"] = database_uri.password.to_s if database_uri.password
+                  add_cli_env_var(vars, "PGHOST", database_uri.host)
+                  add_cli_env_var(vars, "PGPORT", database_uri.port)
+                  add_cli_env_var(vars, "PGUSER", database_uri.user)
+                  add_cli_env_var(vars, "PGPASSWORD", database_uri.password)
                 end
+              end
+
+              def add_cli_env_var(vars, name, value)
+                value = value.to_s
+                return if value.empty?
+
+                vars[name] = value
               end
             end
           end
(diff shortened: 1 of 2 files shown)


lib/hanami/cli/commands/app/db/utils/postgres.rb
```diff
@@ -86,14 +86,21 @@ def escaped_name
...
+              def cli_env_vars
                 @cli_env_vars ||= {}.tap do |vars|
-                  vars["PGHOST"] = database_uri.host.to_s if database_uri.host
-                  vars["PGPORT"] = database_uri.port.to_s if database_uri.port
-                  vars["PGUSER"] = database_uri.user.to_s if database_uri.user
-                  vars["PGPASSWORD"] = database_uri.password.to_s if database_uri.password
+                  add_cli_env_var(vars, "PGHOST", database_uri.host)
+                  add_cli_env_var(vars, "PGPORT", database_uri.port)
+                  add_cli_env_var(vars, "PGUSER", database_uri.user)
+                  add_cli_env_var(vars, "PGPASSWORD", database_uri.password)
```
alassek: This reads awkwardly to me. Let's see if we can make this better:

```ruby
@cli_env_vars ||= %i[host port user password].each_with_object({}) do |field, vars|
  value = database_uri.public_send(field).to_s
  vars["PG#{field}".upcase] = value unless value.empty?
end
```

One of my motivations here is I don't like spreading mutations around when I can keep it in one place.

timriley:
Hi @StantonMatt! I want to thank you for your contribution here, this looks like a good fix. Just one thing I wanted to share: the "validation" list in your PR description reads a little strangely to me. If this is because you're using AI tools to prepare your contributions, can I please ask you to take note of our [AI contribution policy](https://hanakai.org/ai-policy)? Thank you!

StantonMatt:
Thanks for the note, and for linking the policy. Yes, I'm using Codex as an assistant here. I read the policy and will make sure anything I send is reviewed, understood, and in my own voice before it reaches the repo.

I also cleaned up the PR description just now. The old validation list was too mechanical, and one line there did not belong in a project PR description.

PR #417, opened 2026-06-05
Do not create duplicate routes when running the action generator repeatedly

### LINKED ISSUE #146: Action generator appends routes even if exists
When you execute `bundle exec hanami generate action books.index` several times.
You end up with the same routes in routes.rb
Example : 
```ruby
bundle exec hanami generate action books.index
bundle exec hanami generate action books.index

## config/routes.rb
# frozen_string_literal: true

module Bookshelf
  class Routes < Hanami::Routes
    get "/books", to: "books.index"
    get "/books", to: "books.index"
  end
end
``` 

We should have the root only one time.

sandbergja:
Closes #146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 ### Fixed
 
 - Don't overwrite libpq ENV vars with empty strings (@StantonMatt in #414).
+- Skip generating duplicate routes when running `generate action`. (@sandbergja in #417)
 
 ### Security
 
--- a/lib/hanami/cli/commands/app/command.rb
+++ b/lib/hanami/cli/commands/app/command.rb
@@ -44,7 +44,7 @@ module Hanami
               Hanami::Env.load
 
               super
-            rescue FileAlreadyExistsError, RouteAlreadyExistsError => exception
+            rescue FileAlreadyExistsError => exception
               err.puts(exception.message)
               exit(1)
             end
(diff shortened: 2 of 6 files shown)


timriley:
Thanks for working on this, @sandbergja! ❤️ 

Just a thought on the approach: instead of raising `RouteAlreadyExistsError` in `add_route_to_file` and `add_route_to_block` (those are good and well-named new methods, btw!), how about we just turn it into a no-op, i.e. leave the routes file alone and continue execution?

My feeling is that if there really is a legitimate conflict, it would be on the action file itself, which we already handle noisily via the existing `FileAlreadyExistsError`. 

Overall, I would like our generators just "do the right thing" as much as possible without erroring and forcing user reaction, and no-op-ing when a route already exists feels like it fits with this approach. What do you think?

(I also realised while reviewing this that we need to make our action generator aware of our new `resource`-based routing, and most likely prefer that by default. This is a whole separate piece of work, though! I'll make an issue for it: #418)

There's also a bunch of rubocop failures showing in the tests here, and that's because I changed one of our default rules last night. I'll go and fix that separately in the main branch now.

timriley:
RuboCop issues fixed in https://github.com/hanami/hanami-cli/commit/abc244a8838b65bfb3dce3dd36b117aa1243281f, a rebase should sort you out here :)

sandbergja:
Sounds good to me, @timriley, and thanks for the review!  I've rebased and changed the behavior to a no-op (with a short message informing the user that the route is being skipped).

katafrakt:
Just wanted to say: good job! I was trying to solve this issue once, but got stuck somewhere between Hanami CLI and Dry Files. You went into much better direction with the fix.

I also second that conflict should be o no-op, not a blocker - but you already fixed that.

timriley:
Thank you again, @sandbergja, this is great! I just pushed up a changelog entry, and I'll merge this now :)

Issue #400, opened 2026-04-18
Use gem.coop namespaces in generated Gemfile


timriley:
> [!WARNING]
> We can't merge the change described below until our gems have been imported into gem.coop namespaces. I'll leave a comment on this issue (or its respective PR, if someone creates one) once we're ready to go.

If `hanami new` is called with `--gem-source=gem.coop`, then we should use the namespaced versions of our gems.

Something like this:

```ruby
# frozen_string_literal: true

source "https://gem.coop"

source "https://beta.gem.coop/@hanami" do
  gem "hanami", "~> 2.3.0"
  gem "hanami-assets", "~> 2.3.0"
  gem "hanami-controller", "~> 2.3.0"
  gem "hanami-db", "~> 2.3.0"
  gem "hanami-router", "~> 2.3.0"
  gem "hanami-validations", "~> 2.3.0"
  gem "hanami-view", "~> 2.3.0"
end

source "https://beta.gem.coop/@dry" do
  gem "dry-types", "~> 1.7"
  gem "dry-operation", ">= 1.0.1"
end

gem "puma"
gem "rake"
gem "sqlite3"

group :development do
  gem "hanami-webconsole", "~> 2.3.0", source: "https://beta.gem.coop/@hanami"
end

group :development, :test do
  gem "dotenv"
end

group :cli, :development do
  gem "hanami-reloader", "~> 2.3.0", source: "https://beta.gem.coop/@hanami"
end

group :cli, :development, :test do
  gem "hanami-rspec", "~> 2.3.0", source: "https://beta.gem.coop/@hanami"
end

group :test do
  # Database
  gem "database_cleaner-sequel"

  # Web integration
  gem "capybara"
  gem "rack-test"
end
```

Notable changes:

- Grouped all the main Hanami gems under a `source` block (to avoid repetition)
- Did the same for the Dry gems immediately below (also to avoid repetition)
- Used the single-line `source:` option where namespaces gems are used in the various bundler groups further down (it's unlikely that multiple namespaced gems will be needed at these points, so the block form makes less sense)

Later on, we may choose to use these namespaces gems _regardless_ of the top-level gem source, but as a first step, we'll do it when the user explicitly opts into using gem.coop.

timriley:
This is ready to work on now, all our future releases (from Hanami 3.0 onwards) will be published through our @hanami namespace on gem.coop, and all historical gems have been imported.

timriley:
[...]

PR #424, opened 2026-06-08
Source gems from gem.coop namespaces

katafrakt:
With namespaces out of beta, we can now have #400.
Not sure if we need/want changelog entry for that.


timriley (approved):
Perfect, thanks @katafrakt! I do think it's worth a changelog note :)

katafrakt:
Changelog entry added

PR #429, opened 2026-06-25
Provide a way to prevent db command re-runs for test

alassek:
At my workplace, we have a remote development environment named "development" that allows for free-form experimentation with our service infrastructure (yes, this is dumb. No, I can't do anything about it). This environment is isolated from any test systems, so we would like a way to explicitly switch off re-runs when performing `hanami db` commands.

Since we are using this ENV var anyway, there shouldn't be any harm in testing the value explicitly.

### DIFF
--- a/lib/hanami/cli/commands/app/db/command.rb
+++ b/lib/hanami/cli/commands/app/db/command.rb
@@ -179,6 +179,9 @@ module Hanami
               # Re-runs in test are for development-env commands only.
               return unless Hanami.env == :development
 
+              # Re-runs have been intentionally switched off
+              return if ENV.fetch("HANAMI_CLI_DB_COMMAND_RE_RUN_IN_TEST", "true") != "true"
+
               cmd = $0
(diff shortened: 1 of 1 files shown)


timriley:
@alassek Thank you, I'm happy to bring this in, especially if it helps keep your usage of Hanami as clean as possible :)

One caveat though: I think we should leave this undocumented (it's quite niche), and allow ourselves to change in the future. I expect our current "re-run against the test DB" approach to change once we can eventually create a proper API around database interactions in hanami-db. At that point we can think of a cleaner way of disabling the test re-runs.

Does that sound OK to you?

alassek:
@timriley Yeah, leaving this undocumented feels right. I thought about suggesting a `--no-test` flag or something, but I'm not sure that this matters to enough people to be worth making a public API.

timriley:
Now that I think about it, a `--skip-test-db` flag would feel pretty good, and I don't think that would need to change even as our underlying implementation changes.

alassek:
@timriley okay, I'll open a PR

alassek:
#430

PR #430, opened 2026-06-27
Add --skip-test-db flag to DB Commands

alassek:
This is based on the `--env` implementation.


alassek:
The release of rdoc 8.0 today broke jruby: ruby/rdoc#1746

timriley:
Thanks @alassek!

PR #434, opened 2026-07-07
Support SQLite3 JDBC

katafrakt:
Support SQLite3 JDBC adapter.

This is a messy change, because of some differences betreen CRuby and JRuby adapters. Notably:
* JDBC adapter seems to expect the directory for the database file to already exists, CRuby's SQLite3 driver creates it
* `Dir.chdir` is not effective in JRuby for gems calling external JARs and I haven't found an equivalent for JVM - meaning we need to use absolute paths for temporary directories in tests
* Ruby's `URI` does not parse JDBC-compatible URIs well (`jdbc:sqlite:db/dev.sqlite`)

Good news is that I think this is the most messy one and it should be more straightforward for PostgreSQL and MySQL.

### DIFF
--- a/lib/hanami/cli/commands/app/db/utils/sqlite.rb
+++ b/lib/hanami/cli/commands/app/db/utils/sqlite.rb
@@ -67,10 +67,18 @@ module Hanami
               # @api private
               # @since 2.2.0
               def name
-                # Sequel expects sqlite:// URIs to operate the same as file:// URIs: 2 slashes for
-                # a relative path, 3 for an absolute path. In the case of 2 slashes, the first part
-                # of the path is considered by Ruby's `URI` as the `#host`.
-                @name ||= "#{database_uri.host}#{database_uri.path}"
+                @name ||=
+                  if database_uri.scheme == "jdbc"
+                    # For JDBC SQLite URIs like "jdbc:sqlite:db/app.sqlite3",
+                    # we need to extract the path part after "jdbc:sqlite:"
+                    # The standard URI.parse doesn't handle JDBC URIs well, so we remove the prefix manually
+                    database_url.sub(%r{^jdbc:sqlite:}, "")
+                  else
+                    # Sequel expects sqlite:// URIs to operate the same as file:// URIs: 2 slashes for
+                    # a relative path, 3 for an absolute path. In the case of 2 slashes, the first part
+                    # of the path is considered by Ruby's `URI` as the `#host`.
+                    "#{database_uri.host}#{database_uri.path}"
+                  end
               end
 
               private
--- a/lib/hanami/cli/commands/app/db/utils/sqlite.rb
+++ b/lib/hanami/cli/commands/app/db/utils/sqlite.rb
@@ -67,10 +67,17 @@ module Hanami
               # @api private
               # @since 2.2.0
               def name
-                # Sequel expects sqlite:// URIs to operate the same as file:// URIs: 2 slashes for
-                # a relative path, 3 for an absolute path. In the case of 2 slashes, the first part
-                # of the path is considered by Ruby's `URI` as the `#host`.
-                @name ||= "#{database_uri.host}#{database_uri.path}"
+                @name ||= if database_uri.scheme == "jdbc"
+                            # For JDBC SQLite URIs like "jdbc:sqlite:db/app.sqlite3",
+                            # we need to extract the path part after "jdbc:sqlite:"
+                            # The standard URI.parse doesn't handle JDBC URIs well, so we remove the prefix manually
+                            database_url.sub(%r{^jdbc:sqlite:}, "")
+                          else
+                            # Sequel expects sqlite:// URIs to operate the same as file:// URIs: 2 slashes for
+                            # a relative path, 3 for an absolute path. In the case of 2 slashes, the first part
+                            # of the path is considered by Ruby's `URI` as the `#host`.
+                            "#{database_uri.host}#{database_uri.path}"
+                          end
               end
 
               private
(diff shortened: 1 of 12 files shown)


lib/hanami/cli/commands/app/db/utils/sqlite.rb
```diff
@@ -67,10 +67,17 @@ def exec_load_command
...
+                            # For JDBC SQLite URIs like "jdbc:sqlite:db/app.sqlite3",
+                            # we need to extract the path part after "jdbc:sqlite:"
+                            # The standard URI.parse doesn't handle JDBC URIs well, so we remove the prefix manually
+                            database_url.sub(%r{^jdbc:sqlite:}, "")
+                          else
+                            # Sequel expects sqlite:// URIs to operate the same as file:// URIs: 2 slashes for
+                            # a relative path, 3 for an absolute path. In the case of 2 slashes, the first part
+                            # of the path is considered by Ruby's `URI` as the `#host`.
+                            "#{database_uri.host}#{database_uri.path}"
+                          end
```
timriley: Little style nitpick (that I think should now be permitted after recent changes to our rubocop rules) — could we avoid large indentations like this? For something like this the next best thing would be to move the `if` down to the next line and just indent that whole block by 2 spaces only.

lib/hanami/cli/commands/app/db/utils/sqlite.rb
```diff
@@ -67,10 +67,17 @@ def exec_load_command
...
               # @since 2.2.0
               def name
-                # Sequel expects sqlite:// URIs to operate the same as file:// URIs: 2 slashes for
-                # a relative path, 3 for an absolute path. In the case of 2 slashes, the first part
-                # of the path is considered by Ruby's `URI` as the `#host`.
-                @name ||= "#{database_uri.host}#{database_uri.path}"
+                @name ||= if database_uri.scheme == "jdbc"
+                            # For JDBC SQLite URIs like "jdbc:sqlite:db/app.sqlite3",
+                            # we need to extract the path part after "jdbc:sqlite:"
+                            # The standard URI.parse doesn't handle JDBC URIs well, so we remove the prefix manually
```
timriley: Thank you for adding this comment — this is really helpful for future understanding!

spec/support/helpers.rb
```diff
@@ -0,0 +1,34 @@
...
+  module Support
+    module Helpers
+      def sqlite_url(url, dir: nil)
+        url = sqlite_db_name(url, dir:)
+        if jruby?
+          # We need to ensure that the parent directory for the database exists, because JDBC driver
+          # won't create it
+          base_dir = File.dirname(url)
+          FileUtils.mkdir_p(base_dir)
+          "jdbc:sqlite:#{url}"
```
timriley, line 18: This is a fine workaround, but a little surprising from a code perspective — we're adding real side effects to what otherwise feels like a "reader" method.

What would it look like if we took a different approach, e.g. making sure the `db/` directory existed as part of ordinary test setup (maybe built into another helper of some kind) rather than doing it as part of this method? That said, I could understand how this adds a _different_ kind of noise to our tests, so I'm happy to go with your judgement call on this.

spec/unit/hanami/cli/commands/app/db/structure/dump_spec.rb
```diff
@@ -98,8 +98,8 @@ def db_migrate
       SQL
 
       expect(output).to include_in_order(
-        "db/app.sqlite3 structure dumped to config/db/structure.sql",
-        "db/main.sqlite3 structure dumped to slices/main/config/db/structure.sql"
+        "#{sqlite_db_name("db/app.sqlite3", dir: @dir)} structure dumped to config/db/structure.sql",
```
timriley: I think it might be a bit of a downgrade in testing clarity with using this helper method inside our expectations on literal command output.

For example, when I read this, I'm not 100% sure what the output will be for JRuby? Will it be "db/app.sqlite3 structure dumped..." or "/some/full/path/to/db/app.sqlite3 structure dumped..."?

If it's the latter, then I think we really should have some code to strip away the absolute path elements and just show the relative path from the app root.

What would you say about keeping these output expectations as literal strings (since these are basically part of our "user interface"), and then changing other parts of our code to make sure they still output consistently even when running under JRuby?

timriley (commented):
Thank you for doing this, @katafrakt! We're getting closer and closer to Hanami on JRuby, I love it!

I'm happy with the direction here overall, but I did have a couple of small questions that might result in some further change.

I'm happy for you to consider these and then merge once you're ready :) Please feel free to loop me back in if you need, too — I'd be glad to re-review or discuss things further.

spec/support/helpers.rb
katafrakt: I'm glad you called that out. I thought that MRI SQLite driver does that (creates a directory if it does not exist), but I cannot reproduce it now. I need to reanalyze what's really going on here, because in that case we should not have this dir-generating code here.

katafrakt: I removed this and the tests still pass, so that was some kind of a brain fart from my side.

spec/unit/hanami/cli/commands/app/db/structure/dump_spec.rb
katafrakt: I reverted the change in expectations and forced path conversion to relative (if viable) on JRuby: e2695b4b2bb0df9b7521aa7ea53a485b7312ca83 It has a slight downside they if user uses absolute path in configuration the might see relative in the logs, but I think it's acceptable trade-off (?). And alternative would be to have a proper `DatabaseUrl` value object, that keeps information whether the path was relative or absolute in the beginning and only convert based on that - but I think it might be too much.

Let me know what you think @timriley

timriley (approved):
This is great, thanks @katafrakt!

PR #435, opened 2026-07-11
 Run per-slice asset commands in threads instead of forking, for JRuby support

parndt:
`Process.fork` has no JVM equivalent, but each forked child only ran
already-subprocess-spawning code (`InteractiveSystemCall`, backed by
Open3.popen3), so a Thread works identically and is portable. Interrupt
handling now signals the real subprocess pid directly rather than an
intermediate forked Ruby process.

Also avoids ObjectSpace.each_object in the `app_integration` spec
support, which JRuby disables by default, mirroring the same fix in
`hanami/hanami`.


parndt:
Oh amazing!! I just saw #404

PR #436, opened 2026-07-12
fix namespace issue in RubyFileGenerator

mddelk:
Update `RubyFileGenerator` so that it properly handles nested namespaces that might conflict with the app or slice name.

see https://discourse.hanakai.org/t/namespace-issue-when-generating-actions-that-match-the-app-name/1501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 ### Fixed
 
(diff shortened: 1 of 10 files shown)


timriley:
Thank you @mddelk, this is great!!

I've just made a few tweaks (see commits above). The biggest change was removing some of the testing redundancy — for shared behaviour like this I'm trying to set a pattern where we focus more on unit tests and use a smaller set of integration tests to ensure things are exercised as expected.

Merging now! ❤️

PR #437, opened 2026-07-12
Fix psqlrc test issue

mddelk:
Add some postgres test support cleanup.

1. Pass `-X` when calling `psql`, to avoid loading user configuration.

We're using `psql` here to list databases, so we need clean output to parse.

This lets us avoid  issues when `.psqlrc` files are present that result in additional stdout, such as:

```
\pset null '(null)'
```

2. Pass `exception: true` when running postgres `system` commands.

This makes errors here more obvious, e.g:

```
Failures:

  1) Hanami::CLI::Commands::App::DB::Create single db in app postgres creates the database
     Failure/Error: system(cmd_env, "dropdb --force #{database}", exception: true)

     RuntimeError:
       Command failed with exit 1: dropdb --force Null display is "(null)".
     # ./spec/support/postgres.rb:25:in 'Kernel#system'
     # ./spec/support/postgres.rb:25:in 'block (3 levels) in <top (required)>'
     # ./spec/support/postgres.rb:24:in 'Array#each'
     # ./spec/support/postgres.rb:24:in 'block (2 levels) in <top (required)>'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,15 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 ### Changed
 
+- Print clearer errors when Postgres CLI commands fail. (@mddelk in #437)
+
 ### Deprecated
 
 ### Removed
 
 ### Fixed
+
+- Avoid loading user config when running Postgres CLI commands, to ensure command output is in the expected format. (@mddelk in #437)
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


katafrakt:
For reference, as I actually did not know this switch:

```
  -X, --no-psqlrc          do not read startup file (~/.psqlrc)
```

This generally looks like a good change to me.

timriley:
Thanks @mddelk! We're gonna have a hanami-cli patch release soon in your honour :)

PR #441, opened 2026-08-15
Add `hanami new --skip-git` option

aaronmallen:
Generating an app with `--skip-git` skips the `git init` step and the `.gitignore` file, for cases where the app is being generated into an existing repository or is not version controlled at all.


timriley (approved):
Looks good! Don’t forget to add your name and PR number at the end of the CHANGELOG entry (to match the others). 😊

PR #443, opened 2026-08-19
Skip slice view files when hanami-view is not bundled

### LINKED ISSUE #442: Slices generate views even when hanami-view is not installed
## Describe the bug

Calling `bundle exec hanami generate slice <name>` in a hanami application that does not have `hanami-view` installed will still generate view files.

## To Reproduce

1. Create a hanami app with `hanami new <path> --skip-view`
2. Create a slice inside of that hanami app with `hanami generate slice <name>`
3. Observe that a base view class, a views, and a templates directory is generated.

## Expected behavior

I would expect `--skip-view` to skip views

## My environment

- Hanami CLI: v3.0.1
- Ruby version: v4.0.6

aaronmallen:
`hanami generate slice` created `view.rb`, `views/helpers.rb`, `views/context.rb`, and the templates directory unconditionally, so an app generated with `hanami new --skip-view` still got view files in every new slice.

Guard those files behind `Hanami.bundled?("hanami-view")`, matching the check the app, action, and mailer generators already use.

Closes #442

### DIFF
--- a/lib/hanami/cli/generators/app/slice.rb
+++ b/lib/hanami/cli/generators/app/slice.rb
@@ -46,16 +46,6 @@ module Hanami
               auto_register: false
             ).create(force:)
 
-            RubyClassFile.new(
-              fs: fs,
-              inflector: inflector,
-              namespace: slice,
-              key: "view",
-              base_path: directory,
-              parent_class_name: "#{Hanami.app.namespace}::View",
-              auto_register: false
-            ).create(force:)
-
             if Hanami.bundled?("hanami-mailer")
               RubyClassFile.new(
                 fs:, inflector:,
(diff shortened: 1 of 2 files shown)


lib/hanami/cli/generators/app/slice.rb
```diff
@@ -182,6 +142,55 @@ def call(app, slice, url, force: false, **opts) # rubocop:disable Metrics/AbcSiz
 
           attr_reader :inflector
 
+          def generate_view?
+            Hanami.bundled?("hanami-view")
+          end
```
katafrakt: Not sure about such one-line helpers, especially given that the rest of conditional generation uses `Hanami.bundled?` directly. So even for the sake of consistency, I'd skip it.

katafrakt (approved):

PR #446, opened 2026-09-05
Display relative path to SQLite DB on JRuby

katafrakt:
The last minute change in #434 was incorrect. The database path was changed to relative in `database.rb`, but `sqlite.rb` overrode it, displaying absolute path and causing 46 specs to fail unnecessarily. Also, I somehow sneaked in `Dir.current` call, which does not exist - switching to `Dir.pwd` instead.

### DIFF
--- a/lib/hanami/cli/commands/app/db/utils/database.rb
+++ b/lib/hanami/cli/commands/app/db/utils/database.rb
@@ -79,15 +79,7 @@ module Hanami
               end
 
               def name
-                path = database_path
-                # On JRuby we probably converted a relative path to an absolute path, because
-                # the JDBC driver only accepts the latter. Now it's time to convert it back.
-                # This means that JRuby users might sometimes see relative path when they actually
-                # used abosulte path in the configuration.
-                if jruby?
-                  pathname = Pathname.new(path).expand_path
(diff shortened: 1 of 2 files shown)


lib/hanami/cli/commands/app/db/utils/database.rb
```diff
@@ -79,15 +79,7 @@ def initialize(slice:, gateway_name:, system_call:)
...
-                path = database_path
-                # On JRuby we probably converted a relative path to an absolute path, because
-                # the JDBC driver only accepts the latter. Now it's time to convert it back.
-                # This means that JRuby users might sometimes see relative path when they actually
-                # used abosulte path in the configuration.
-                if jruby?
-                  pathname = Pathname.new(path).expand_path
-                  path = pathname.relative_path_from(Dir.current).to_s if pathname.to_s.start_with?("#{Dir.current}/")
-                end
+                path = relativize_for_display(database_path)
```
katafrakt: Actually, do we even need it here? This relative vs absolute path is a concern only for SQLite, so probably the change should be only there.

katafrakt: changed according to this comment

timriley: Thanks @katafrakt! Happy with your call :)

